PR #4, opened 2016-08-03
Handle unrecognized commands

smarterclayton:
Now that Dockerfile syntax is changing again, tolerate unrecognized
commands by default.


smarterclayton:
@vbatts

PR #27, opened 2016-12-22
imageprogress: document goroutine leak

mmilata:
Not 100% sure here but it seems to me that when using the `imageProgress.Write` method, [this for loop](https://github.com/openshift/imagebuilder/blob/86214f339d010b2df2ceb8729f95d3ad2090690b/imageprogress/progress.go#L178) is exited only on error (which may not happen) or on EOF which doesn't happen because the `internalWriter` pipe end is never closed. This results in the goroutine started in `Write` hanging in `decoder.Decode`.

I don't know how to fix it but figured documenting the behavior might save someone some time.

### DIFF
--- a/imageprogress/pull.go
+++ b/imageprogress/pull.go
@@ -9,6 +9,8 @@ import (
 // on pull progress of a Docker image. It only reports when the state of the
 // different layers has changed and uses time thresholds to limit the
 // rate of the reports.
+//
+// WARNING: may leak a goroutine, do not use in long-running processes.
 func NewPullWriter(printFn func(string)) io.Writer {
 	return newWriter(pullReporter(printFn), pullLayersChanged)
 }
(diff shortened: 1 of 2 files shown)


smarterclayton:
The context image progress is used in is underneath Executor.PullImage, which calls into `go-dockerclient` to do the pull, which has full timeout and close support.  I ran a few scenarios and don't see a goroutine leaked (because go-dockerclient sends an EOF when the server finishes responding).  If there was a leak, the bug would be in go-dockerclient for not properly finishing output to OutputStream.

mmilata:
I was thinking that the Reader part of io.Pipe() result [returns EOF](https://golang.org/pkg/io/#PipeReader.Read) only when the corresponding Writer is closed, which `go-dockerclient` doesn't do.

Let me close this since it's a nonproblem in realistic scenarios.

Issue #17, opened 2016-11-16
VOLUME contains data added after declaration


twaugh:
I understand this is a known issue, that a Dockerfile like the following will end up creating a volume differently when built using imagebuilder:
```
FROM busybox
RUN mkdir /vol
RUN touch /vol/foo
VOLUME /vol
RUN touch /vol/bar
```

It may be that a volume is carefully set up in a Dockerfile, then some other command is run which -- as a side effect -- modifies that volume content. The author may be using the knowledge that the volume declaration 'freezes' the initial volume content in order to know the side-effect-producing command is safe to run.

I'd like to propose a solution: during the implementation of VOLUME, imagebuilder could export the content of the container and capture the content. This is the author's intended initial content for the volume.

After all the Dockerfile instructions are run, a final set of instructions could restore the content of volumes declared in the Dockerfile to the state they were in at declaration. For instance, saving the current user, adding a binary to clean out volume content (and afterwards, itself), becoming root and running it, then restoring the current user.

smarterclayton:
This is a known issue, there's really no way to fix it without executing things in the container, which we can only do if we know that the image has a shell in it.  So we can only fix part of the problem.

smarterclayton:
How critical is this?

twaugh:
Injecting a static `rm` binary was the proposal I had.

Short of that, if imagebuilder could at least warn when the volume content ends up different than at declaration time, it would at least enable any tooling using it to re-build the container image a different way if it wants full compatibility for VOLUME.

smarterclayton:
Mostly fixed, but see #30 for the gap with owner permissions.  I will add a `--strict-volume-ownership` flag to prevent this accidentally being lost.

smarterclayton:
Added a check in #31 which will fail if RUN follows VOLUME.  ADD following VOLUME is successful.

PR #28, opened 2016-12-24
Handle volume layering

smarterclayton:
Any Dockerfile with a VOLUME must be two layers - one for the base
contents and one for the volume definitions. Add support for this and a
test.

First part of fixing #17


smarterclayton:
Important note - if Docker sees:

```
ADD file /var
VOLUME /var
ADD file2 /var
```

the final output will be:

```
/var/file
/var/file2
```

It's only RUN commands that don't get added.  This PR adds the framework and testing without actually changing existing behavior, but sets a bunch of preparation up for that PR.

PR #37, opened 2017-03-13
bump(github.com/fsouza/go-dockerclient):a45f2c476754603041aeb54751207d5e5ec30b77

smarterclayton:
Jump to the version selected by @bparees for openshift/origin#13235

I pruned some things from godeps that were in your branch - go-dockerclient/testing, which removed gorilla. Also, golang.org/x/sys/* were not necessary, or github.com/Azure or github.com/Microsoft (we might have broader dependencies, but noted that).


bparees:
thought we couldn't do this because of k8s?  if you're able to bump the main vendored fsouza, then you might as well cherry-pick my main commit from https://github.com/openshift/origin/pull/13235 too and we'll be all set on the networking front.

smarterclayton:
This is the imagebuilder repo.

bparees:
[sigh]

PR #45, opened 2017-05-31
cleanup goroutine properly in imageprogress.NewPullWriter

jim-minter:
- imageProgressWriter should be a WriteCloser, so that a Close() will allow it to clean up its internal goroutine.
- removed error propagation logic because I believe it's not used anywhere, and it's not feasible under the Writer interface contract.
- removed `func (w *imageProgressWriter) ReadFrom` because it's also broken and not actually necessary.
- changed layerStatus from map[string]progressLine to a map[string]*progressLine to save a (small) copy.

### DIFF
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -6,6 +6,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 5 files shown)


jim-minter:
@smarterclayton ptal

jim-minter:
@smarterclayton ptal

PR #49, opened 2017-08-22
Update github.com/sirupsen/logrus, things which depend on it, and things they require

nalind:
This update vendors in a newer version of `github.com/sirupsen/logrus`, newer versions of the dependencies we vendor which used it under the old name so that they'll use it by the new name, and other packages which they depend on.  It then adapts to a couple of API changes in newer versions of `github.com/docker/docker`.


nalind:
I had some difficulty with godep, so this may well fail CI in a dramatic fashion.

nalind:
@smarterclayton, PTAL

smarterclayton:
This adds 300k lines of code, I think there are new dependencies that slipped in.  For instance, grpc, gen-foto, golang/crypto, go-memdb, coreos/etcd/raft.  You may have accidentally grabbed something too broad, or we may need to snip a dependency in the core libraries.

nalind:
A large number of these come by way of the updated `go-dockerclient`, which pulls in `github.com/docker/docker/opts`, which now uses `github.com/docker/docker/daemon/cluster/convert`, which among other things pulls in `github.com/docker/swarmkit/api`, which uses `github.com/coreos/etcd/raft/raftpb`, and `github.com/docker/swarmkit/ca`, which pulls in `github.com/hashicorp/go-memdb`, `github.com/cloudflare/cfssl`, and `golang.org/x/crypto`.

I managed to pare out a few things that we didn't end up needing, and have uploaded what should be the full inverted dependency tree to https://gist.github.com/nalind/ab98c50af24c066c66de2609f5a72716 if it's of any help here - I don't see any build tags that we could use which would cut this down, so if you have any advice on how to snip things, I'm happy to look into them.

rhatdan:
@smarterclayton Any chance you can look at this, this is preventing us from updating buildah with newer versions of containers/image and containers/storage.

smarterclayton:
The latest version of fsouza/go-dockerclient depends on a lot less if you pair it with the latest docker core.  Is that possible?

nalind:
Indeed it does!  Updated.

rhatdan:
@smarterclayton PTAL

smarterclayton:
Tested locally, thanks.

PR #50, opened 2017-08-24
Handle VOLUME statements referring to missing dirs

nalind:
When a VOLUME statement refers to a directory which isn't present in the container (for example, in the Dockerfile that's part of the Docker source tree), our attempts to save and restore its contents will fail.  When attempting to save, detect the 404 error we get for that case and note that the directory should be empty.  When restoring, don't bother trying to extract a snapshot of such a directory, just clean it out.

### DIFF
--- a/dockerclient/conformance_test.go
+++ b/dockerclient/conformance_test.go
@@ -349,6 +349,10 @@ func TestConformanceInternal(t *testing.T) {
 			Name:       "noworkdir",
 			Dockerfile: "testdata/Dockerfile.noworkdir",
 		},
+		{
+			Name:       "volumeexists",
+			Dockerfile: "testdata/Dockerfile.volumeexists",
+		},
 	}
 
 	for i, test := range testCases {
--- a/dockerclient/testdata/Dockerfile.volumeexists
+++ b/dockerclient/testdata/Dockerfile.volumeexists
@@ -1,0 +1,5 @@
+FROM busybox
+RUN mkdir -p 0700 /var/lib/bespoke-directory
+RUN chown 1:1 /var/lib/bespoke-directory
+VOLUME /var/lib/bespoke-directory
+RUN touch /var/lib/bespoke-directory/emptyfile


smarterclayton:
Can you add a conformance test for this?

nalind:
Okay, I've added a conformance test.  I also reworked the patch so that it issues a `mkdir -p` before attempting to snapshot the directory and waits for unmonitored commands that it issues to finish executing, which together should make the conformance test pass, something which the original proposed change didn't actually accomplish.

dockerclient/client.go
```diff
@@ -833,6 +833,45 @@ func snapshotPath(path, containerID string, client *docker.Client) (string, erro
...
 	glog.V(4).Infof("Snapshot %s for later use under %s", path, f.Name())
 
+	ensurePath := func(dest string) error {
+		if !strings.HasSuffix(dest, "/") {
+			dest = dest + "/"
+		}
+		glog.V(4).Infof("Running mkdir -p %q", dest)
+		exec, err := client.CreateExec(docker.CreateExecOptions{
+			Container: containerID,
+			Cmd:       []string{"/bin/sh", "-c", "mkdir -p $@", "", dest},
```
smarterclayton: This will fail on images that don't have SH like scratch images.  Won't a tar file successfully create these?

nalind: It won't be created as a side-effect if the layer doesn't include any content to put inside of it.

smarterclayton: This makes images without SH unbuildable though - so we can't do this as is.  What other options do we have?

nalind: Okay, uploading an empty directory seems to work, and doesn't require us to exec anything.  I've updated the patch to try doing it that way.

nalind:
Switched to uploading an archive containing the volume directory if, at the point it's declared, we can't read its contents.

rhatdan:
@nalind @smarterclayton Is this PR still active?

nalind:
The 'volumerun' test in the tree and the 'novolume' test that we add here fail without the code change from this PR, so I think so.

dockerclient/client.go
```diff
@@ -609,6 +610,53 @@ func (e *ClientExecutor) Preserve(path string) error {
...
+		if !strings.HasSuffix(dest, "/") {
+			dest = dest + "/"
+		}
+		err := e.Client.DownloadFromContainer(e.Container.ID, docker.DownloadFromContainerOptions{
+			Path:         dest,
+			OutputStream: ioutil.Discard,
+		})
+		return err
+	}
+	if err := readPath(path); err != nil {
```
smarterclayton: Why is the check necessary?  Can't we just always upload the dir and rely on later stages to fail (for instance, RUN would fail because cd would fail).

rhatdan:
@nalind is this still needed?

dockerclient/client.go
nalind: If the volume directory already existed in the container, this would overwrite its ownership unconditionally.  I'll add a conformance test that fails if the take this bit out.

sosiouxme:
The final commit with the new conformance test merges fine and should IMHO still be included now that #113 merged. Otherwise, this PR is already completed.
/lgtm
@rhatdan  ?

sosiouxme:
oh, i can lgtm stuff after all :)

PR #52, opened 2017-11-08
Guard against invalid progress data from Docker

csrwng:
Fixes https://github.com/openshift/origin/issues/13041


csrwng:
@smarterclayton ptal

PR #58, opened 2018-01-23
Add support for passing build args via --build-arg flag

mumoshu:
build args had been almost there but a command-line flag to actually pass it to `imagebuilder`.
This commit adds the flag so that `imagebuilder` supports `--build-arg` which populates a value for `ARG` in `Dockerfile`.

Note that there seems like a bit difference between `docker-build` and `imagebuilder` regarding `--build-arg`:

- docker-build is known to leave the build args used while building inside the image metadata
- imagebuilder doesn't leave the build args there

I'm not sure why `imagebuilder`'s implementation doesn't leave build args in image metadata but it seems like a slightly preferable behavior.

`docker-build`'s build args are often used to pass secret variables to a docker-bulid context, which ends up leaving the secrets inside the docker image.
`imagebuilder` users won't be trapped like this due to the different behavior.

```
$ docker build --build-arg foo=bar1 -t buildargtester1 .
Sending build context to Docker daemon  17.63MB
Step 1/4 : FROM alpine:3.7
 ---> 3fd9065eaf02
Step 2/4 : ARG foo
 ---> Using cache
 ---> 6c16d009e365
Step 3/4 : RUN echo $foo > foo.out
 ---> Running in 242096e61485
 ---> 6ef29b90049e
Removing intermediate container 242096e61485
Step 4/4 : RUN cat foo.out
 ---> Running in 2d3895b73aa9
bar1
 ---> 2d76605b424a
Removing intermediate container 2d3895b73aa9
Successfully built 2d76605b424a
Successfully tagged buildargtester1:latest
```

```
$ docker run -it --rm buildargtester1 cat foo.out
bar1
```

```
$ docker inspect buildargtester1 | grep bar1
                "foo=bar1",
```

```
$ ./myimagebuilder --build-arg foo=bar2 -t buildargtester2 .
--> FROM alpine:3.7
--> ARG foo
--> RUN echo $foo > foo.out
--> RUN cat foo.out
bar2
--> Committing changes to buildargtester2 ...
--> Done
```

```
$ docker run -it --rm buildargtester2 cat foo.out
bar2
```

```
$ docker inspect buildargtester2 | grep bar2
```

### DIFF
--- a/cmd/imagebuilder/imagebuilder.go
+++ b/cmd/imagebuilder/imagebuilder.go
@@ -28,7 +28,7 @@ func main() {
 
 	flag.Var(&tags, "t", "The name to assign this image, if any. May be specified multiple times.")
 	flag.Var(&tags, "tag", "The name to assign this image, if any. May be specified multiple times.")
-	flag.Var(&arguments, "build-arg", "An optional list of build-time variables usable as ARG in Dockerfile. Use --build-arg ARG1=VAL1 --build-arg ARG2=VAL2 syntax for passing mutliple build args.")
+	flag.Var(&arguments, "build-arg", "An optional list of build-time variables usable as ARG in Dockerfile. Use --build-arg ARG1=VAL1 --build-arg ARG2=VAL2 syntax for passing multiple build args.")
 	flag.StringVar(&dockerfilePath, "f", dockerfilePath, "An optional path to a Dockerfile to use. You may pass multiple docker files using the operating system delimiter.")
 	flag.StringVar(&dockerfilePath, "file", dockerfilePath, "An optional path to a Dockerfile to use. You may pass multiple docker files using the operating system delimiter.")
 	flag.StringVar(&imageFrom, "from", imageFrom, "An optional FROM to use instead of the one in the Dockerfile.")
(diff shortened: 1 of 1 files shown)


mumoshu:
Hi @smarterclayton, thanks a lot for maintaining this great project 👍 

I came to know this project while searching for a way to build a small, secure docker images fast.
I'm contributing this change as `--build-arg` is the only missing feature for me.
Would you mind reviewing this? Thanks!

cmd/imagebuilder/imagebuilder.go
```diff
@@ -24,8 +24,11 @@ func main() {
 	var imageFrom string
 	var mountSpecs stringSliceFlag
 
+	arguments := stringMapFlag{}
+
 	flag.Var(&tags, "t", "The name to assign this image, if any. May be specified multiple times.")
 	flag.Var(&tags, "tag", "The name to assign this image, if any. May be specified multiple times.")
+	flag.Var(&arguments, "build-arg", "An optional list of build-time variables usable as ARG in Dockerfile. Use --build-arg ARG1=VAL1 --build-arg ARG2=VAL2 syntax for passing mutliple build args.")
```
smarterclayton: Type `multiple`

smarterclayton:
One comment, then this is LGTM

mumoshu:
@smarterclayton Thanks for reviewing! Fixed it.

PR #59, opened 2018-02-22
Support multi-stage image builds

smarterclayton:
Adds support for the `FROM foo AS bar` and `COPY --from=bar` syntax,
    allowing multi-stage builds to be executed. The `-target` flag is supported
    on `cmd/imagebuilder` to allow a specific stage to be chosen.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/imagebuilder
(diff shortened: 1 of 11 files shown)


smarterclayton:
@bparees closes the gap

smarterclayton:
Did some manual testing, going to add conformance tests soon.

PR #60, opened 2018-02-27
STOPSIGNAL needs to be cross-platform

smarterclayton:
We don't actually process the signal, we just verify that we recognize
it, and imagebuilder can work across platform. Skip verifying the signal
value and just verify the name is recognized.


smarterclayton:
@dcbw not related to the networking_minimal failure, just how image builder works across platforms

PR #63, opened 2018-03-27
Allow (capital) AS in FROM statement

bertinatto:
This solves an issue created in projectatomic/buildah#538.


rhatdan:
LGTM

bertinatto:
@smarterclayton, something like this?

PR #69, opened 2018-04-20
add NoAuthFn to make imagebuilder vendorable

mfojtik:
@smarterclayton we need this otherwise you will never be able to initialize the AuthFn outside this repo (vendored types).


smarterclayton:
Ugh, I think we need to define our own AuthConfig.

smarterclayton:
We need to make the public interfaces usable in general (via either a type or interface).

Issue #71, opened 2018-04-23
Multi-stage build doesn't reuse previous filesystem correctly


smarterclayton:
Running a multi stage build like:

```
FROM centos:7 AS base
RUN touch /1

FROM base
RUN ls /1
```

should work.  Instead, `base` is interpreted as a local image.

smarterclayton:
Fixed by #72

PR #72, opened 2018-04-24
When building multiple stages, use a previous container

smarterclayton:
Later stages of a multi-stage build should use previous stage containers
when the FROM matches an AS.


smarterclayton:
Fixing #71

PR #73, opened 2018-05-02
initial RPM packaging

sosiouxme:
I cargo-culted stuff just to get this working. Is there a more simplified template to follow?

### DIFF
--- a/imagebuilder.spec
+++ b/imagebuilder.spec
@@ -1,0 +1,63 @@
+
+#debuginfo not supported with Go
+%global debug_package %{nil}
+
+# modifying the Go binaries breaks the DWARF debugging
+%global __os_install_post %{_rpmconfigdir}/brp-compress
+
+%{!?commit: %global commit HEAD }
+
+#
+# Customize from here.
+#
+
+%global golang_version 1.8.1
+%{!?version: %global version 0.0.1}
+%{!?release: %global release 1}
+%global package_name imagebuilder
+%global product_name Container Image Builder
+%global import_path github.com/openshift/imagebuilder
+
+Name:           %{package_name}
+Version:        %{version}
+Release:        %{release}%{?dist}
+Summary:        Builds Dockerfile using the Docker client
+License:        ASL 2.0
+URL:            https://%{import_path}
+
+Source0:        https://%{import_path}/archive/%{commit}/%{name}-%{version}.tar.gz
+BuildRequires:  golang >= %{golang_version}
+
+### AUTO-BUNDLED-GEN-ENTRY-POINT
+
+%description
+Builds Dockerfile using the Docker client
+
+%prep
+GOPATH=$RPM_BUILD_DIR/go
+rm -rf $GOPATH
+mkdir -p $GOPATH/{src/github.com/openshift,bin,pkg}
+%setup -q -c -n imagebuilder
+cd ..
+mv imagebuilder $GOPATH/src/github.com/openshift/imagebuilder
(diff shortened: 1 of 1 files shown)


sosiouxme:
I got it to build in brew.

sosiouxme:
@smarterclayton is this ok or would you like further refinement?

Need a brew built RPM for OSBS to use.

smarterclayton:
If all you need is the spec file, I'd recommend just going the simple route
and strip this down to just a spec file, and just run go install in it.
Nothing about this tool really requires the extra ceremony.

smarterclayton:
You shouldn't even need any of the cross arch stuff except to run GOARCH=X during your `go install`.  Also we don't want tito.

sosiouxme:
> If all you need is the spec file, I'd recommend just going the simple route
> and strip this down to just a spec file, and just run go install in it.
>

I tried that, and it was just not obvious how to make it work. I probably
only needed some env vars set but well, packaging not my strength and
everyone else is busy.

We do want builds in multiple arches, but I suppose that would be handled
by running builds on multiple architectures...

I literally just copied code from
https://github.com/openshift/image-registry/ -- no one would point me at
anything simpler.

I can poke at it some more though to try to simplify.

smarterclayton:
I would remove:

* hack/
* .tito/

Comments inline for spec

imagebuilder.spec
```diff
@@ -0,0 +1,129 @@
...
+%prep
+%if 0%{do_prep}
+%setup -q
+%endif
+
+%build
+%if 0%{do_build}
+%if 0%{make_redistributable}
+# Create Binaries for all internally defined arches
+%{os_git_vars} hack/build-cross.sh
```
smarterclayton: Remove all this (and any os_git_vars reference), and remove the make_redistributable check.

In the if arch sections, replace the entire line with `GOOS=<value_of_first_part> GOARCH=<value_of_second_part> go install ./cmd/imagebuilder`

Then in the install step you'll need to move `GOPATH/bin/imagebuilder` into the right location.

imagebuilder.spec
```diff
@@ -0,0 +1,129 @@
...
+%ifarch ppc64le
+  BUILD_PLATFORM="linux/ppc64le"
+%endif
+%ifarch %{arm} aarch64
+  BUILD_PLATFORM="linux/arm64"
+%endif
+%ifarch s390x
+  BUILD_PLATFORM="linux/s390x"
+%endif
+OS_ONLY_BUILD_PLATFORMS="${BUILD_PLATFORM}" %{os_git_vars} hack/build-cross.sh
```
smarterclayton: This would just be `go install ./cmd/imagebuilder`

imagebuilder.spec
```diff
@@ -0,0 +1,129 @@
...
+%install
+
+PLATFORM="$(go env GOHOSTOS)/$(go env GOHOSTARCH)"
+install -d %{buildroot}%{_bindir}
+
+# Install linux components
+for bin in imagebuilder
+do
+  echo "+++ INSTALLING ${bin}"
+  install -p -m 755 _output/local/bin/${PLATFORM}/${bin} %{buildroot}%{_bindir}/${bin}
```
smarterclayton: This is the GOPATH/bin dir.

sosiouxme:
How about this then? Just the spec file.

The trick was just to get everything in GOPATH.

sosiouxme:
@smarterclayton poke about this

smarterclayton:
Good enough for government work

PR #74, opened 2018-05-18
Handle copy destination with single dot properly

jpeeler:
A copy operation with a destination of "." was not taking into account the
working directory properly, resulting in the working dir becoming the
destination. Essentially when a destination is ".", the working directory is
what needs to be checked for proper prefixes and suffixes.

For reference the dockerfile snippet I was testing with looked like:

...
FROM scratch
WORKDIR /root/
COPY --from=builder /go/src/github.com/jpeeler/podpreset-crd/controller-manager .
ENTRYPOINT ["./controller-manager"]
CMD ["--install-crds=false"]

### DIFF
--- a/dispatchers_test.go
+++ b/dispatchers_test.go
@@ -1,0 +1,35 @@
+package imagebuilder
+
+import (
+	"reflect"
+	"testing"
+
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+func TestDispatchCopy(t *testing.T) {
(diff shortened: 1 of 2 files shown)


smarterclayton:
Can you add unit tests please?  Appreciate the fix.

smarterclayton:
Specifically add a conformance test and a copy_test since both of these are an endless source of regressions.

PR #80, opened 2018-06-20
Handle non-trailing slashes in COPY code correctly

### LINKED ISSUE #79: COPY path_without_slash . when working dir is root does not work
```
FROM x
WORKDIR /
COPY directory .
```

where directory contains `/etc/file` does not put `file` into `/etc`, but into `/directory/etc/file`.

Broke openshift ansible.

smarterclayton:
Fixed three cases:

1. Copying a file to root
2. Copying a directory (without a trailing slash) to root
3. Copying a file as a tar to another path

and added more tests.

Fixes #79


smarterclayton:
@sosiouxme significant changes, this is going to go into 3.10 but you may want to test them out at broader scale.

sosiouxme:
Cool, thanks, probably fixes a number of builds.

I've found something similar, will raise another issue.

PR #81, opened 2018-06-24
`COPY --from=` did not properly handle wildcards

smarterclayton:
Some cases where a wildcard was specified would fail.  Refactored the core code to share code paths and more carefully processed directory paths.  New manual test cases created (future change will turn these into a conformance suite).


smarterclayton:
@sosiouxme see this and the test suite pushed after it

PR #90, opened 2018-07-24
Switch to use parsers from github.com/moby/buildkit

rhatdan:
This should give us better handling of multi line labels
and environment variables.


rhatdan:
This PR includes a large vendor created by 
dep ensure
To pull in the buildkit bits.

smarterclayton:
Please split the vendor changes into their own commit.  I'll take a look at this soon, although if this is going to cause problems when vendored into origin we need to know that too before we merge.

rhatdan:
Sadly this PR did not solve the issue of multi-line labels when I vendored it into buildah.

rhatdan:
Since this did not actually fix the issue, and the issue I saw failed the same way on Docker, I am closing this.

PR #91, opened 2018-07-24
Ensure job control is on in interactive shells

### LINKED ISSUE #88: imagebuilder waits for child processes to complete; docker build doesn't
With a Dockerfile that runs a simple bash script which spawns a child, I found that docker build would finish and kill the build container as soon as the script finished, while imagebuilder would wait for the child to finish:

```
$ git clone https://github.com/sosiouxme/scratch -b 20180720-subprocess-wait

$ docker build --no-cache scratch
Sending build context to Docker daemon 72.19 kB
Step 1/3 : FROM fedora:28
 ---> 75aeb7f897fd
Step 2/3 : ADD script.sh /usr/bin/
 ---> 12ff6dbff87c
Removing intermediate container 64a93a2cbce6
Step 3/3 : RUN script.sh
 ---> Running in fae77c14c89b
parent done
 ---> ab6d22ef095c
Removing intermediate container fae77c14c89b
Successfully built ab6d22ef095c

$ imagebuilder scratch
--> FROM fedora:28 as 0
--> ADD script.sh /usr/bin/
--> RUN script.sh
parent done
child done
--> Committing changes ...
--> Done
```

This might not seem like a big problem but it is the source of my hung builds.

In the logging-kibana build, as part of the build, the node process starts up a daemon. This is because it runs some slow optimize/cache code up front, to store the result in the image so that it doesn't have to be done at runtime. But it doesn't properly kill the daemon so imagebuilder waits forever for it to finish.

Arguably this is just a badly-behaving build, just like if it started an infinite loop in the foreground. But infinite loops are more evident than forgotten children. If folks are testing their builds with docker they will not consider this a problem and imagebuilder will seem broken. And in our production OSBS, at least, the build never completes, blocking any other attempts at building until an admin can cancel it.

smarterclayton:
Without TTY on exec, shell processes behave inconsistently with docker
build. A script that forks will wait to exit in imagebuilder but not
docker build unless we also set our input to a TTY.

Fixes #88

### DIFF
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -685,6 +685,7 @@ func (e *ClientExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 	exec, err := e.Client.CreateExec(docker.CreateExecOptions{
 		Cmd:          config.Cmd,
 		Container:    e.Container.ID,
+		Tty:          true,
 		AttachStdout: true,
 		AttachStderr: true,
 		User:         config.User,
--- a/dockerclient/testdata/runchild/Dockerfile
+++ b/dockerclient/testdata/runchild/Dockerfile
@@ -1,0 +1,4 @@
+FROM busybox
+ADD child.sh .
+RUN /child.sh
+RUN ps -ef
--- a/dockerclient/testdata/runchild/child.sh
+++ b/dockerclient/testdata/runchild/child.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+echo "starting child"
+( sleep 10; echo "child done" ) &
+echo "script done"


rhatdan:
LGTM

smarterclayton:
@sosiouxme can you test this out locally and verify it fixes the break for you?

sosiouxme:
@smarterclayton yes, LGTM.

smarterclayton:
ok, adding a test before i merge.

smarterclayton:
Ugh, using a TTY causes all the container output to change to be `\r\n` instead of `\n`

smarterclayton:
TTY won't work - docker does not create a TTY for build RUN so we'd just be inconsistent.

It looks like the fundamental different is that docker build RUN is doing ContainerAttachRaw, while we're doing Exec.  ContainerAttachRaw does not open stdin when called from build.  The exec code path is not opening stdin either, but it's possible that down in the guts of the process there's something slightly different about how the file descriptors are hooked up.

smarterclayton:
There's a chance this is a bug in exec in docker 1.13.

rhatdan:
I would turn to @nalind for help on this.

smarterclayton:
We still need to figure out what docker build does with regards to job control.

PR #95, opened 2018-09-10
Add field for handling AS clause in FROM

TomSweeneyRedHat:

When processing a Dockerfile line like:

FROM centos AS base

Imagebuilder was catching this, but was not storing the "AS" portion away so the caller could reference it.  We're now noting the value in the Builder struct so Buildah an others can reference it.

I wasn't quite sure how to test this here, if tests are desired, I'll have to noodle a bit more.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -224,6 +224,8 @@ type Builder struct {
 	PendingCopies  []Copy
 
 	Warnings []string
+
+	AsImageName string
 }
 
 func NewBuilder(args map[string]string) *Builder {
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -187,6 +187,9 @@ func from(b *Builder, args []string, attributes map[string]bool, flagArgs []stri
 			return fmt.Errorf("Windows does not support FROM scratch")
 		}
 	}
+	if len(args) == 3 {
+		b.AsImageName = args[2]
+	}
 	b.RunConfig.Image = name
 	// TODO: handle onbuild
 	return nil


TomSweeneyRedHat:
@rhatdan @nalind PTAL

rhatdan:
LGTM
@smarterclayton PTAL

builder.go
```diff
@@ -224,6 +224,8 @@ type Builder struct {
 	PendingCopies  []Copy
 
 	Warnings []string
+
+	AsImageName []string
```
smarterclayton: Why is this an array?

smarterclayton: An individual builder is invoked for a given stage - perhaps it would be cleaner if you make this an executor responsibility (like dockerclient)?

TomSweeneyRedHat: cut/paste error, it should be a string.

TomSweeneyRedHat: I've repushed as a string for now, ty for spotting that.  

My first pass through I changed the name field to be 'name + " AS " + args[2] which eventually set the b.RunConfing.Image  to that value.  The executor then sniffed out the " AS " in the Image and handled it from there.  I thought that was a little funky and was concerned that it might mess up other folks using imagebuilder, so I went to this current approach which I thought made handling the AS part of stuff a bit cleaner, but did suffer the weight of the new field in the struct.

Had you other thoughts for how to get the executor to handle the AS clause?  Wouldn't be the first time I missed something obvious!

dispatchers.go
```diff
@@ -187,6 +187,9 @@ func from(b *Builder, args []string, attributes map[string]bool, flagArgs []stri
 			return fmt.Errorf("Windows does not support FROM scratch")
 		}
 	}
+	if len(args) ==3 {
```
smarterclayton: Please always use gofmt in your editor and format on save.  We don't have the bot on for this repo yet, but it would be required in almost any of them.

TomSweeneyRedHat:
For such a relatively small diff PR, I certainly found ways to mess it up.  Sorry about the gofmt, I'm usually good about that.  I'll look into getting that added into travis for the project next week.  I usually use just base vi and don't like to add on to it, I'm a bit of a dinosaur.

I'll try to answer what was going on.  Prior to these changes, if the Dockerfile had something like:

FROM centos AS base
RUN stuff
RUN more stuff
FROM base
RUN some other stuff

The first stage would have the image name of Centos, but the " AS base" part of that was recognized by the case statement at line 177 in dispatchers.go as being valid, but nothing else was done with it, it was just dropped.  When Buildah worked on the stage it would show it's progress as: 
Step 1: FROM centos

and it had no clue that "base" was associated with that FROM clause.  Then when Buildah worked on stage 2 and had the "base" from the "FROM base" line, it would try to look up the "base" image on the registries and then locally and would fail.

With this change in place, Buildah now checks the AsImageName field which is populated for the first stage.  When it finds a value in that field, it notes the value and when the builder object representing the image from the first FROM is completed, it tucks it away into a map.  Then if a future stage has a name that's in that map, Buildah grabs that builder object from the map and uses it to complete the build.

If you, @nalind or others think there's a better way to approach this, I'm happy to do so.  If you want the details from the Buildah side, the PR in waiting there is: https://github.com/projectatomic/buildah/pull/998

TomSweeneyRedHat:
@smarterclayton @nalind any further thoughts/suggestions or can we merge this?
@rhatdan fyi

smarterclayton:
I’m asking why does buildah not work at the stages level (Ie the []Stage array created for working with multistage builds)?

TomSweeneyRedHat:
@smarterclayton I took another look, I need to test more in the morning, but I think I can make this happen strictly on the Buildah side without changes here.  Buildah does use the stages.  I missed that the stage name is either a number value in a string if there is no " AS " in the FROM command in the Dockerfile.  If there is an " AS " then that stage name is set to the value after the AS in that line.

Buildah could check that stage name and set that value aside if it wasn't numeric, but I'm always leery chopping up a string to see if it's contents are numeric.  I think it would be cleaner to check the AsImageName value as set in this PR.  

So this approach is cleaner from the Buildah side IMO, but if you're concerned this will upset the apple cart in some way for others I can adjust.  

I'll work up a PR in Buildah to verify I can do it just on that side and if you prefer, we can go with that approach.

smarterclayton:
Let’s step back.

The AS field is the name of the stage and can implicitly be used in
subsequent FROM statements.  If not set by a user, it is set as a number
string.  You must always use the stage name (numeric or not) as a reference
to the previous stage once it has appeared.  I don’t understand why you are
concerned about checking for a numeric value, because you must treat the
stage name as a string no matter what.

Maybe you can describe or point to some example code in buildah that shows
how you intend to use the stage name.  I don’t understand how you are
attempting to use the stage name differently from how imagebuilder uses it
today, because the stage executor only needs stage name as an entry into a
map for COPY.


On Sep 21, 2018, at 1:13 AM, Tom Sweeney <notifications@github.com> wrote:

@smarterclayton <https://github.com/smarterclayton> I took another look, I
need to test more in the morning, but I think I can make this happen
strictly on the Buildah side without changes here. Buildah does use the
stages. I missed that the stage name is either a number value in a string
if there is no " AS " in the FROM command in the Dockerfile. If there is an
" AS " then that stage name is set to the value after the AS in that line.

Buildah could check that stage name and set that value aside if it wasn't
numeric, but I'm always leery chopping up a string to see if it's contents
are numeric. I think it would be cleaner to check the AsImageName value as
set in this PR.

So this approach is cleaner from the Buildah side IMO, but if you're
concerned this will upset the apple cart in some way for others I can
adjust.

I'll work up a PR in Buildah to verify I can do it just on that side and if
you prefer, we can go with that approach.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/imagebuilder/pull/95#issuecomment-423363568>,
or mute the thread

TomSweeneyRedHat:
The stage name is as you stated.  From past scars in other projects, I'm just leery about checking to see if a string is full of numbers or other types of characters.  I'd rather have a field that I could reference with a value knowing that I didn't have to do a regex or AtoI type thing to determine contents and then base logic on the findings of those contents.

I went ahead today and created another PR in Buildah that doesn't look for the field that I created, I'm driving it off the stage name.  Given that I"m going to close this one out.  Thanks for the discussion.

PR #96, opened 2018-09-11
WIP - Revert getError removal from PR 45

TomSweeneyRedHat:
I'm adding back the error handling that was removed vrom #45 in order to hopefully address https://bugzilla.redhat.com/show_bug.cgi?id=1626228.

However the progress_test.go tests that should error out on error don't seem to be:

```
# go test
--- FAIL: TestReports (0.04s)
    --- FAIL: TestReports/detect_error (0.01s)
    	progress_test.go:95: detect error: did not get expected error
--- FAIL: TestErrorOnCopy (0.00s)
	progress_test.go:125: Did not get an error when copying to writer
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x513e20]

goroutine 19 [running]:
testing.tRunner.func1(0xc4200805a0)
	/usr/lib/golang/src/testing/testing.go:711 +0x2d2
panic(0x53cc60, 0x61fba0)
	/usr/lib/golang/src/runtime/panic.go:491 +0x283
github.com/openshift/imagebuilder/imageprogress.TestErrorOnCopy(0xc4200805a0)
	/root/tsweeney/workspaces/imagebuilder/src/github.com/openshift/imagebuilder/imageprogress/progress_test.go:127 +0x280
testing.tRunner(0xc4200805a0, 0x56f1c0)
	/usr/lib/golang/src/testing/testing.go:746 +0xd0
created by testing.(*T).Run
	/usr/lib/golang/src/testing/testing.go:789 +0x2de
exit status 2
FAIL	github.com/openshift/imagebuilder/imageprogress	0.053s
```

I'm not sure if they're failing as I've not set up the error condition in the test properly or if the change from #45 of io.Writer to io.Writer.Closer is causing the error not to be generated/caught.  As I've not looked at this code prior to this afternoon, any thoughts/suggestions welcomed.

### DIFF
--- a/imageprogress/progress.go
+++ b/imageprogress/progress.go
@@ -3,6 +3,7 @@ package imageprogress
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"regexp"
@@ -192,15 +193,23 @@ func (w *imageProgressWriter) readProgress(decoder *json.Decoder) error {
 		if err != nil {
 			return err
 		}
-		w.processLine(line)
+		err = w.processLine(line)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
 
-func (w *imageProgressWriter) processLine(line *progressLine) {
+func (w *imageProgressWriter) processLine(line *progressLine) error {
+
+	if err := getError(line); err != nil {
+		return err
+	}
+
 	// determine if it's a line we want to process
 	if !islayerStatus(line) {
-		return
+		return nil
 	}
 
 	w.layerStatus[line.ID] = line
@@ -208,7 +217,7 @@ func (w *imageProgressWriter) processLine(line *progressLine) {
 	// if the number of layers has not stabilized yet, return and wait for more
 	// progress
 	if !w.isStableLayerCount() {
-		return
+		return nil
 	}
 
 	r := createReport(w.layerStatus)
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
@runcom @rhatdan @mtrmac PTAL

mrunalp:
@bparees ptal

bparees:
the error handling itself looks reasonable to me, can't speak to the unit test issue

mtrmac:
I’ll take a detailed look in a bit, but I don’t think it’s quite _this_ simple.

Consider the usual status flow, with an error arriving as the last update:
- The writer calls `imageProgressWriter.Write("the error"})
- `w.internalWriter.Write()` succeeds writing it to the pipe, simultaneously `w.readProgress()` succeeds reading it from the pipe, and decoding the error
- The reading goroutine now has the `err != nil` value we want to record, and calls `pipeIn.CloseWithError()`
- _If_, at this point, the writer sent another data chunk, it would probably receive the error and it would be propagated correctly. But because the error tends to be the last item, this is fairly unlikely (and depends on how exactly the data is chunked/buffered)
- More likely, the writing side would call `imageProgressWriter.Close`, and the `CloseWithError` should cause `imageProgressWriter.Close` to return this error value…
- … but nothing guarantees that `imageProgressWriter.Close` is only called _after_ `pipeIn.CloseWithError` AFAICS; the two can race with each other.
- … but, also, callers of `imageProgressWriter` just do `defer pullWriter.Close` and throw away the error value anyway.

It seems to me that we will need to use a channel to reliably get the error from the reading/decoding goroutine to the writer on `imageProgressWriter.Close`, and then all users will need to check the value.

mtrmac:
> It seems to me that we will need to use a channel to reliably get the error from the reading/decoding goroutine to the writer on `imageProgressWriter.Close`, and then all users will need to check the value.

https://github.com/openshift/imagebuilder/pull/97

PR #98, opened 2018-09-13
Add codeowners file

TomSweeneyRedHat:

Adds the initial CODEOWNERS file.  @smarterclayton and @bparees if you'd like to be added too, happy to do so.


bparees:
@TomSweeneyRedHat i meant this: https://github.com/openshift/origin/blob/master/OWNERS

bparees:
which is what prow consumes

smarterclayton:
Yeah, we don't use codeowners.

PR #101, opened 2018-09-19
Copy .dockerignore into image for compatibility with Docker

### LINKED ISSUE #100: Imagebuilder doesn't copy .dockerignore into the image
https://github.com/openshift/imagebuilder/blob/fbd4f31c568b767447302e08fb3531b33bcbf837/dockerclient/client.go#L136

This is inconsistent with docker. I'm trying to add imagebuilder as an option in our CI to validate we work with it before getting our component into OSBS, but part of our jobs validates that there's no unstaged changes after re-generating some files (to ensure commited generated files look the same in CI as what the developer produced), but this causes issues with that since the .dockerignore is missing in the image after building.

chancez:
From what I can tell, this is inconsistent with how `docker build`. I went through `git blame`, but it ended at the first commit where this is extracted out from Openshift by clayton, so I didn't dig deeper to see if there's a good reason for excluding this file.

Fixes #100

### DIFF
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -129,12 +129,9 @@ func NewClientExecutor(client *docker.Client) *ClientExecutor {
 }
 
 func (e *ClientExecutor) DefaultExcludes() error {
-	excludes, err := imagebuilder.ParseDockerignore(e.Directory)
-	if err != nil {
-		return err
-	}
-	e.Excludes = append(excludes, ".dockerignore")
-	return nil
+	var err error
+	e.Excludes, err = imagebuilder.ParseDockerignore(e.Directory)
+	return err
 }
 
 // WithName creates a new child executor that will be used whenever a COPY statement


smarterclayton:
Locally you can run the conformance tests to verify this didn't break anything (or a test) with `make test-conformance` - if you've run that let me know.

rhatdan:
@smarterclayton @nalind Does this look ok to you?

smarterclayton:
We really need to get the conformance test running as part of the PR (it's mostly blocked because we need a VM to get a docker socket) - until then @nalind or @rhatdan if you can run the conformance suite on this PR against 1.13 so we have confidence that it doesn't regress that suite that would help.

bparees:
@adambkaplan another .dockerignore related buildah item...

chancez:
Hitting this in origin CI also

rhatdan:
LGTM

rhatdan:
I had @nalind review, and I am going to merge, to match Docker behaviour.

PR #107, opened 2018-10-26
Allow container runtime team to be Approvers of imagebuilder

rhatdan:

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,6 +1,8 @@
 reviewers:
-- nalind
 - bparees
-- rhatdan
 approvers:
+- TomSweeneyRedHat
+- mrunalp
+- nalind
+- rhatdan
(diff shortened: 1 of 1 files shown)


rhatdan:
@smarterclayton PTAL

smarterclayton:
Please be aware that image builder is used extensively through our tooling infrastructure.  Untested changes can break openshift CI instantly.  Since conformance tests are manual right now, please enforce that everyone run them before merging.

I would probably say that we need to split this repo into two parts - one for imagebuilder core library, and one for imagebuilder docker client.  The core library is the one that the buildah team needs to iterate on.  The docker client is the one that can break lots of people if changed.  Splitting the core packages and moving generic code from dockerclient to that location is going to be safer than iterating here.

PR #109, opened 2018-11-10
Allow relative paths in `COPY --from=STAGE` by checking working dir

smarterclayton:
Use the previous stage's working dir to do relative paths.


smarterclayton:
This is needed in 4.0 and 3.11 in order to build some of the multi-stage docker files in use by the team. #110 is the backport

Issue #108, opened 2018-10-28
ProcessWord{,s} removes quotes


vrothberg:
When using buildah, the content of specified labels will be altered in a way that specified quotes get removed. This can lead to various kinds of bugs and undesired side-effects.

If, for instance, a Dockerfile specifies `LABEL RUN /usr/bin/foo --string-arg "a string arg"` removing the quotes is likely to cause issues when the image is being used in special contexts (e.g., `podman container runlabel`).

This issue affects Podman and Buildah. Note that Docker has the very same issue, which I consider to be a bug for the reasons mentioned above. I had a look at the code and it seems fixable but I am not overly familiar with it and want to collect some feedback before.

Cc @baude @mheon @TomSweeneyRedHat @rhatdan 

@baude, we could reuse the functions for runlabel's env replacements.

rhatdan:
Yes this could be a big problem.

vrothberg:
@rhatdan, can we get this fixed? It's blocking progress for `podman-container-runlabel`.

rhatdan:
Sure.  PR's welcomed. :^)

vrothberg:
> Sure. PR's welcomed. :^)

Just added it to my todo ;-)

vrothberg:
Starting to work on this now (Cc @baude)

vrothberg:
Closing as described in #111.

PR #111, opened 2018-11-14
ProcessWords: preserve quotes

vrothberg:
Preserve single and double quotes in ProcessWords to match the
function description and user intention.  Also add some simple
unit tests to catch potential future regressions.

Fixes: #108
Signed-off-by: Valentin Rothberg <vrothberg@suse.com>

### DIFF
--- a/builder_test.go
+++ b/builder_test.go
@@ -222,7 +222,7 @@ func TestBuilder(t *testing.T) {
 			Dockerfile: "dockerclient/testdata/Dockerfile.env",
 			From:       "busybox",
 			Config: docker.Config{
-				Env:   []string{"name=value", "name2=value2a            value2b", "name1=value1", "name3=value3a\\n\"value3b\"", "name4=value4a\\\\nvalue4b"},
+				Env:   []string{"name=value", "name2=\"value2a            value2b\"", "name1=value1", "name3=\"value3a\\n\"value3b\"\"", "name4=\"value4a\\\\nvalue4b\""},
 				Image: "busybox",
 			},
 		},
@@ -245,7 +245,7 @@ func TestBuilder(t *testing.T) {
 			Config: docker.Config{
 				User:         "docker:root",
 				ExposedPorts: map[docker.Port]struct{}{"6000/tcp": {}, "3000/tcp": {}, "9000/tcp": {}, "5000/tcp": {}},
-				Env:          []string{"SCUBA=1 DUBA 3"},
+				Env:          []string{"SCUBA=\"1 DUBA 3\""},
 				Cmd:          []string{"/bin/sh", "-c", "echo 'test' | wc -"},
 				Image:        "busybox",
 				Volumes:      map[string]struct{}{"/test2": {}, "/test3": {}, "/test": {}},
@@ -327,8 +327,8 @@ func TestBuilder(t *testing.T) {
 				},
 			},
 			Config: docker.Config{
-				Env:    []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin", "FOO=value"},
-				Labels: map[string]string{"test": "value"},
+				Env:    []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin", "FOO=\"value\""},
+				Labels: map[string]string{"test": "\"\"value\"\""},
 			},
 		},
 		{
@@ -337,8 +337,8 @@ func TestBuilder(t *testing.T) {
 			From:       "busybox",
 			Config: docker.Config{
 				Image:  "busybox",
-				Env:    []string{"FOO=value", "TEST=", "BAZ=first"},
-				Labels: map[string]string{"test": "value"},
+				Env:    []string{"FOO=\"value\"", "TEST=", "BAZ=first"},
+				Labels: map[string]string{"test": "\"\"value\"\""},
 			},
 			Runs: []Run{
 				{Shell: true, Args: []string{"echo $BAR"}},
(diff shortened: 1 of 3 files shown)


vrothberg:
@rhatdan PTAL. Need Yoda to have a look at the changes as it smells a bit of some side-effects.

TomSweeneyRedHat:
@vrothberg do you have a Dockerfile that you could attach here or in the issue that shows the problem before the fix and now works with the fix?  I saw the example Dockerfile line, but would like an easy to use Dockerfile if you've one handy.

thaJeztah:
I wonder if this is a bug, or if the intent was to follow shell quoting rules here (this could be if the same rules are applied to `ENV`, and people expecting it to be processed more in a "shell-like" matter, i.e.;

```bash
echo "beta" gamma 'delta'
beta gamma delta
```

If not, then this could be a bug, where only leading/trailing quotes should be removed from values, but it's applying that rule incorrectly.

The pitfall there being that there are two syntaxes;

```Dockerfile
LABEL name value
LABEL name=value name=value
```

Using this Dockerfile;

```Dockerfile
FROM scratch

# Sets label 'one'
LABEL one two three four

# Sets label 'five' and 'seven'
LABEL five=six seven=eight

# Sets label 'alpha'
LABEL alpha "beta" gamma 'delta'

# Sets label 'epsilon' and 'eta'
LABEL epsilon='zeta' eta="theta"

# Sets label 'quoted'
LABEL quoted "a quoted \"value\""

# Sets label 'quoted2'
LABEL quoted2="another quoted \"value\""
```

Building the above currently produces;

```
{
  "alpha": "beta gamma delta",
  "epsilon": "zeta",
  "eta": "theta",
  "five": "six",
  "one": "two three four",
  "quoted": "a quoted \"value\"",
  "quoted2": "another quoted \"value\"",
  "seven": "eight"
}
```

But if the _intent_ was to remove leading/trailing quotes, I _think_ the expected results would be:

```json
{
  "alpha": "\"beta\" gamma 'delta'",
  "epsilon": "zeta",
  "eta": "theta",
  "five": "six",
  "one": "two three four",
  "quoted": "a quoted \"value\"",
  "quoted2": "another quoted \"value\"",
  "seven": "eight"
}
```

thaJeztah:
ping @tonistiigi @akihirosuda @duglin PTAL

thaJeztah:
I found this earlier discussion around this; https://github.com/moby/moby/issues/36027#issuecomment-358194966 (also https://github.com/moby/moby/issues/32140#issuecomment-289918601(

thaJeztah:
From those discussions (and the docs change in https://github.com/docker/cli/pull/814), it looks like the _intent_ was to process it with shell rules (and thus quotes to be stripped);

> The value will be interpreted for other environment variables, so quote characters will be removed if they are not escape

vrothberg:
@thaJeztah, thanks a lot for the very valuable input!

Your comments made my mind shift to the question whether the values of LABEL must be processed the same way as the ones of ENV. My personal preference would be to preserve quotes and to not transform the input for labels, but given it's been like that for such a long while I fear there's nothing to change without breaking existing workloads.

For the use case of `podman-container-runlabel`, we might need to mention this explicitly in the docs to avoid this pitfall.

thaJeztah:
Perhaps @justincormack's proposal in https://github.com/moby/moby/issues/32140#issuecomment-289465543 would be a solution for this; add a JSON syntax for `LABEL` (and `ENV`) that won't substitute variables, e.g.;

```Dockerfile
LABEL {"RUN":"/usr/bin/foo --string-arg \"a string arg\""}
```

or

```Dockerfile
LABEL ["RUN=/usr/bin/foo --string-arg \"a string arg\""]
```

Pitfalls there are (although less problematic than for the `RUN` syntax, which is ambiguous), not everyone knows that JSON doesn't support _single quotes_ (see https://github.com/moby/moby/issues/5701, and https://github.com/moby/moby/pull/7998)

duglin:
The intent was to preserve quotes like the linux shell, however, since the value args all get squashed into one string things get a little weird. Look at this:

```
$ cat a.sh
echo $*

$ echo ads "asd asd" asd | xargs ./a.sh
ads asd asd asd
```
Notice no quotes.

I would point out that:
`LABEL RUN2 "/usr/bin/foo --string-arg 'a string arg'" `
does give you want you're looking for.
`"RUN2": "/usr/bin/foo --string-arg 'a string arg'",`

So, if you want to keep the quotes you may need to wrap the whole thing with more quotes, or escape things:

For example:
`ENV foo3 asd \"asd qwe\" qwe`
yields:
"foo3=asd \"asd qwe\" qwe",

Which does seem to make some sense to me. But I'm not a sh expert to know for certain.

rhatdan:
I think we do what the user expects.  I would expect that LABEL or a RUN command to work the same as if I typed in the command at the shell.

Currently we have the users jumping through hoops to make this work.  So I like the idea of this patch.

The issue will be what happens if the programmer did jump through hoops will we break that use case.

Does
ENV foo3 asd \"asd qwe\" qwe
end up with
"foo3=asd \"asd qwe\" qwe",

vrothberg:
> Does
ENV foo3 asd "asd qwe" qwe
end up with
"foo3=asd "asd qwe" qwe",

No. It ends put with `"foo3=asd asd qwe qwe"` (i.e., quotes removed).

vrothberg:
I think we can really break it down to if LABEL should be processed in a similar fashion as ENV.

ENTRYPOINT, for instance, is processed differently (for the reasons mentioned above) and quotes are properly escaped: `"/usr/bin/foo --string-arg \"a string arg\""`

thaJeztah:
The difference there is that `ENTRYPOINT` is used as argument for `/bin/sh -c`, and to be evaluated by `/bin/sh` at _runtime_, whereas `LABEL` is processed/evaluated during build.


Given this Dockerfile;

```Dockerfile
FROM scratch
ENTRYPOINT /usr/bin/foo --string-arg "a string arg"
```

The entrypoint will be;

```bash
docker inspect --format '{{json .Config.Entrypoint}}' myimage
["/bin/sh","-c","/usr/bin/foo --string-arg \"a string arg\""]
```

thaJeztah:
Irregardless of the above; changing this would be quite a breaking change, and potentially breaking a lot of existing Dockerfiles

vrothberg:
> Irregardless of the above; changing this would be quite a breaking change, and potentially breaking a lot of existing Dockerfiles

That's what I am afraid of as well.

vrothberg:
I am closing this PR as I changing this behavior might cause regressions for some users. Quotes need to be escaped manually and we need to educate users to be aware of that.

PR #112, opened 2018-12-18
Bump "safe" dependencies

vdemeester:
This bump some "safe" dependencies atomically (as 1 update per commit). By "safe" dependency, I means it tried to update dependency that are not constrained, nor overridden, and that do not track `master`.

This has been done using [`sec`](https://github.com/vdemeester/sec), which means:
- It tried to update the dependency
- Ran tests to validate it doesn't break (at least the build) (`go test ./...` or `go test ./pkg/...`, …).
- Commit if the test command was successful.

Note: this *never* update to `latest` either (if `latest` is the latest release, even if it's a major version bump). It's only updating minor or fixes versions.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -61,12 +61,12 @@
   revision = "b68221c37ee597950364788204546f9c9d0e46a1"
 
 [[projects]]
-  digest = "1:87dcb59127512b84097086504c16595cf8fef35b9e0bfca565dfc06e198158d7"
+  digest = "1:ade935c55cd6d0367c843b109b09c9d748b1982952031414740750fdf94747eb"
   name = "github.com/docker/go-connections"
   packages = ["nat"]
   pruneopts = "NUT"
-  revision = "3ede32e2033de7505e6500d6c868c2b9ed9f169d"
-  version = "v0.3.0"
+  revision = "7395e3f8aa162843a74ed6d48e79627d9792ac55"
+  version = "v0.4.0"
 
 [[projects]]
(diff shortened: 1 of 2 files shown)


rhatdan:
@mrunalp @bparees @TomSweeneyRedHat  PTAL

vendor/github.com/Microsoft/go-winio/pipe.go
```diff
@@ -133,12 +138,14 @@ func (s pipeAddress) String() string {
...
-// takes longer than the specified duration. If timeout is nil, then the timeout
-// is the default timeout established by the pipe server.
+// takes longer than the specified duration. If timeout is nil, then we use
+// a default timeout of 5 seconds.  (We do not use WaitNamedPipe.)
 func DialPipe(path string, timeout *time.Duration) (net.Conn, error) {
 	var absTimeout time.Time
 	if timeout != nil {
 		absTimeout = time.Now().Add(*timeout)
+	} else {
+		absTimeout = time.Now().Add(time.Second * 2)
```
TomSweeneyRedHat: I'm assuming this is from upstream, but the comment says 5 seconds and it looks like it's only using 2 seconds?

vdemeester: yeah this is from upstream :sweat_smile:

TomSweeneyRedHat:
LGTM and Buildah passes its baseline tests with this in place.  Would like a head nod from @nalind or @mrunalp

PR #113, opened 2019-01-01
 issue 87: ensure WORKDIR exists

sosiouxme:
Building on https://github.com/openshift/imagebuilder/pull/50 this PR also ensures that the WORKDIR exists when specified.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -40,6 +40,7 @@ type Run struct {
 
 type Executor interface {
 	Preserve(path string) error
+	EnsureContainerPath(path string) error
 	Copy(excludes []string, copies ...Copy) error
 	Run(run Run, config docker.Config) error
 	UnrecognizedInstruction(step *Step) error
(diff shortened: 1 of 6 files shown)


rhatdan:
/approve
@nalind PTAL

rhatdan:
@TomSweeneyRedHat PTAL

TomSweeneyRedHat:
@smarterclayton had a comment in #50 and that code is included in this PR.  That should be addressed and I'd feel more comfortable holding this one until @nalind gets back.  It LGTM otherwise though.

sosiouxme:
@TomSweeneyRedHat any idea when @nalind gets back?
I needed this fix internally so used this branch to build an RPM... would really like these problems solved in master.

TomSweeneyRedHat:
@nalind is just back and probably still looking through the backlog of emails.  @nalind PTAL

sosiouxme:
thanks!

PR #119, opened 2019-01-17
ADD supports the chown option also

rhatdan:
We need to support the --chown operator for ADD as well as COPY.

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -128,9 +128,20 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 	if len(args) < 2 {
 		return errAtLeastOneArgument("ADD")
 	}
+	var chown string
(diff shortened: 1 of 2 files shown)


rhatdan:
@smarterclayton @nalind @TomSweeneyRedHat PTAL

rhatdan:
@bparees PTAL

rhatdan:
Fixes  https://bugzilla.redhat.com/show_bug.cgi?id=1534505

TomSweeneyRedHat:
Should you add a test similar to https://github.com/openshift/imagebuilder/blob/9e710ba9bcf926e020306b67633bf38b8e671da5/dispatchers_test.go#L38 ?

TomSweeneyRedHat:
FWIW, LGTM too.  Thanks for adding the test.

PR #122, opened 2019-02-19
Make the comment in docker history slightly more readable

smarterclayton:
```
docker history ac6a8767ddfb
IMAGE               CREATED             CREATED BY          SIZE                COMMENT
ac6a8767ddfb        5 seconds ago       #(imagebuilder)
sleep 86400         0B
d8233ab899d4        4 days ago          /bin/sh -c #(nop)  CMD ["sh"]                   0B
<missing>           4 days ago          /bin/sh -c #(nop) ADD file:9ce77bda0ecf8a7f5…   1.2MB
```

Also bump golang.org/x/sys, does not compile on MacOS due to unix.Major and unix.Minor missing


smarterclayton:
@ecordell

PR #127, opened 2019-03-16
dockerfile: remove unused files

vrothberg:
Remove unused source files from dockerfile as imagebuilder requires only
the parser and command sub-package.

Partial fix for: https://github.com/openshift/imagebuilder/pull/125#issuecomment-473562682

Signed-off-by: Valentin Rothberg <rothberg@redhat.com>


vrothberg:
@rhatdan @TomSweeneyRedHat @siretart PTAL

siretart:
While trying to compile this patch, I get this compilation error:

```
src/github.com/openshift/imagebuilder/dockerclient/archive.go:416:15: undefined: idtools.IDPair                                                                                                                    
dh_auto_build: cd obj-x86_64-linux-gnu && go install -gcflags=all=\"-trimpath=/<<BUILDDIR>>/golang-github-openshift-imagebuilder-1.0\+git20190308.705fe92/obj-x86_64-linux-gnu/src\" -asmflags=all=\"-trimpath=/<<\
BUILDDIR>>/golang-github-openshift-imagebuilder-1.0\+git20190308.705fe92/obj-x86_64-linux-gnu/src\" -v -p 4 github.com/openshift/imagebuilder github.com/openshift/imagebuilder/cmd/imagebuilder github.com/opensh\
ift/imagebuilder/dockerclient github.com/openshift/imagebuilder/dockerfile/command github.com/openshift/imagebuilder/dockerfile/parser github.com/openshift/imagebuilder/dockerfile/parser/dumper github.com/opens\
hift/imagebuilder/imageprogress github.com/openshift/imagebuilder/signal github.com/openshift/imagebuilder/strslice returned exit code 2                                                                           
make: *** [debian/rules:8: build] Error 2                                                                                                                                                                          
```

Do I need to update some dependency? If so, which one?

vrothberg:
> Do I need to update some dependency? If so, which one?

The idtools package has changed as well. github.com/containers/storage/pkg/idtools supports the old API of this package, but @nalind or @TomSweeneyRedHat or @rhatdan will know for sure.

We're not hitting this issue in buildah (although we use the latest version of docker) as the failing package is not vendored in.

I assume you're trying to compile imagebuilder with another version of Docker than the vendored one, right?

siretart:
{quote}
I assume you're trying to compile imagebuilder with another version of Docker than the vendored one, right?
{quote}

Quite possible. In https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=924257#34 one of the maintainers of the `docker.io` source package has asked me to upload docker.io version 18.09.3. I strongly suspect that this corresponds to https://github.com/docker/docker-ce/tree/v18.09.3 which in turn leads to https://github.com/docker/engine/tree/18.09/pkg/idtools. https://github.com/docker/engine/blob/18.09/pkg/idtools/idtools.go does not appear to contain the IDPair structure.

siretart:
Valentin, I believe with "new" API, you refer to this commit: 
https://github.com/docker/engine/commit/763d8392612942ff5c32a35f8bdafd7ae93d3321#diff-afbdc8affebc599d929dfe3eb8b7de71R105

Quoting from the commit message:
> This implements chown support on Windows. Built-in accounts as well
> as accounts included in the SAM database of the container are supported.
> 
> NOTE: IDPair is now named Identity and IDMappings is now named
> IdentityMapping.

Yeah, in Debian I don't think linking / bundling an older version of docker is an option.  Any ideas how to work around that?

The most irritating part seems to be that they not only renamed "IDPair" to "Identity", but also made it incompatibly by adding an additional member "SID" into it (probably Windows support related). Maybe the fork can avoid that type?

elboulangero:
@siretart said:

> In https://salsa.debian.org/docker-team/docker/blob/experimental/debian/upstream-version-gitcommits I see that this release corresponds to git commit 774a1f4. I strongly suspect that this corresponds to https://github.com/docker/docker-ce/tree/v18.09.3

Correct, that's what we package in debian experimental at the moment.

vrothberg:
> Yeah, in Debian I don't think linking / bundling an older version of docker is an option. Any ideas how to work around that?

Can you check the last update? We now vendor containers/storage and the packages there instead of the ones from docker/docker.

rhatdan:
LGTM

siretart:
With git revision `d08ca8de8ca0a66b3e2bb00af438b8b330096c98`, I am now (finally) able to build the dockerfile image package. @vrothberg  thanks a lot for these efforts, I think this PR is good to merge.  

FYI, my packaging work can be found here: https://salsa.debian.org/go-team/packages/golang-github-openshift-imagebuilder/tree/master/debian - note that I've added this PR as a distro patch. Ideally I wouldn't have to do that in the future.

Valentin, just to clarify, the dependency hierarchy looks like this:

  container/image -> openshift/imagebuilder -> container/storage

Did i get that right?

I still need to upload the new docker.io package to experimental for @elboulangero which I promise to do soon after my final test builds pass. This is a prerequisite for this package build.

vrothberg:
Hi Reinhard,

sorry for the late reply.

> With git revision d08ca8de8ca0a66b3e2bb00af438b8b330096c98, I am now (finally) able to build the dockerfile image package.

That's great news. Thanks a lot for bearing with us.

>Valentin, just to clarify, the dependency hierarchy looks like this:

> container/image -> openshift/imagebuilder -> container/storage

I think the first one is containers/**buildah**, right? If so, the hierarchy looks fine. @rhatdan did a wonderful job removing the cycle between image & storage so we are getting closer.

The example from above with containers/image may look like this:
```
buildah -> imagebuilder -> storage
         |               |   
         -----> image ----
```

Note that we are also about to move the versioning scheme to of the libraries to semver. We currently do not have plans to support stable version of the libraries but the usage of [go mod](https://github.com/golang/go/wiki/Modules) is slowly but continuously growing in the community which requires semver.

vrothberg:
CI is green, I think we're good to merge. @rhatdan @TomSweeneyRedHat @nalind PTAL.

TomSweeneyRedHat:
LGTM, but would like a head nod from @rhatdan or @nalind

TomSweeneyRedHat:
btb, very nice work @vrothberg !

PR #129, opened 2019-05-21
Use imported constants instead of our own

nalind:
We can get `commandArg` and `commandStopSignal` from the updated import, so we don't need to define them ourselves any more.


TomSweeneyRedHat:
LGTM assuming happy tests.

rhatdan:
LGTM

nalind:
Rebased.

PR #130, opened 2019-07-12
Migrate from glog to klog

adambkaplan:
glog is no longer being actively maintained.
Move to klog so that we play nicely with downstream projects that want to use klog (ex openshift/builder).

/assign @smarterclayton 

/cc @vrothberg @nalind


adambkaplan:
@smarterclayton looks like this repo hasn't been updated to fix the travis issue

/skip travis-ci/pr

PR #133, opened 2019-07-25
Bug 1730446: dockerclient newArchiveMapper: always end archiveRoot with "/"

nalind:
Always end the `archiveRoot` value that we supply as the `Path` for a `DownloadFromContainer` request with a `/`, so that if it refers to a symbolic link, the engine knows we want an archive of the directory that it points to, rather than an archive containing just the symlink.


rhatdan:
LGTM
@TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM

PR #134, opened 2019-07-26
[openshift-3.11] Bug 1730446: dockerclient newArchiveMapper: always end archiveRoot with "/"

nalind:
Always end the `archiveRoot` value that we supply as the `Path` for a `DownloadFromContainer` request with a `/`, so that if it refers to a symbolic link, the engine knows we want an archive of the directory that it points to, rather than an archive containing just the symlink.

Cherry-picked from #133, so that we can merge this into origin's 3.11 branch.

### DIFF
--- a/dockerclient/archive.go
+++ b/dockerclient/archive.go
@@ -357,6 +357,9 @@ func newArchiveMapper(src, dst string, excludes []string, resetOwners bool, chec
 			prefix = path.Base(archiveRoot)
 		}
 	}
+	if !strings.HasSuffix(archiveRoot, "/") {
+		archiveRoot += "/"
+	}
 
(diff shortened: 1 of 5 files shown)


nalind:
We've got two Travis indicators, one of which isn't getting updated.
/override continuous-integration/travis-ci/pr

nalind:
It doesn't look like I can force the merge from here.  @bparees, is this something you're able to help us with?

bparees:
i manually added the approved label, not sure why /approve isn't working since you guys are all in the OWNERS file.

nalind:
I don't think we have an OWNERS on the target branch.

bparees:
@nalind if you want to add one i'll get it merged.

nalind:
@bparees, that'd be great!  Opened #136.

PR #142, opened 2019-10-10
Add automatic platform args in the global scope

grooverdan:
Based on https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

By default assumed build platform = target platform as builder doesn't [yet?] gave platform.


TomSweeneyRedHat:
In Docker, this is only supported starting with v18.09 while using BuildKit.  The changes here look reasonable to me, @nalind WDYT?

rhatdan:
@mrunalp WDYT?

grooverdan:
Any objections @nalind, @mrunalp ?

grooverdan:
So this is missing some /lgtm and removing the /size/xxl :heart: 
/joike

grooverdan:
:traffic_light:

/joke

grooverdan:
Can you push the vendor github.com/containerd/containerd updates separately? Then I can remove that from the series. That will remove the size/XXL tag enabling this to be merged.

TomSweeneyRedHat:
@nalind @mrunalp @redhatdan PTAL

TomSweeneyRedHat:
And scrolling back I realize I never LGTM'd.  So doing so now.
LGTM

PR #145, opened 2019-10-17
Add env to stage

QiWang19:
Add env from containers.conf to stage so buildah bud can set default env for containers using containers.conf.

PR buildah reads containers.conf https://github.com/containers/buildah/pull/1858

Signed-off-by: Qi Wang <qiwan@redhat.com>

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -212,6 +212,7 @@ func NewStages(node *parser.Node, b *Builder) (Stages, error) {
 			Builder: &Builder{
 				Args:        b.Args,
 				AllowedArgs: b.AllowedArgs,
+				Env:         b.Env,
 			},
 			Node: root,
 		})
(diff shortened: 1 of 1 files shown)


QiWang19:
@rhatdan PTAL

rhatdan:
/approve
lgtm
@nalind @TomSweeneyRedHat @vrothberg @giuseppe PTAL

TomSweeneyRedHat:
Can we add a test to this to make sure the right value gets set?
The code otherwise LGTM

rhatdan:
@TomSweeneyRedHat I am not sure there is a way to test this now, we have been testing it within buildah though.

QiWang19:
> Can we add a test to this to make sure the right value gets set?
> The code otherwise LGTM

I think this should be tested in buildah reading the containers.conf

TomSweeneyRedHat:
@QiWang19 and @redhatdan thx for the 411, I'll take the hold off then.

Issue #137, opened 2019-08-20
COPY --from doesn't fail when source file doesn't exist


rcerven:
when copying file from previous stage
via COPY --from=previous_stage
build doesn't fail on COPY command when source file doesn't exist

--> COPY --from=stage1 /opt/non-existent  /file_from_stage1
--> RUN cat /file_from_stage1
cat: /file_from_stage1: No such file or directory
running 'cat /file_from_stage1' failed with exit code 1

nalind:
This should be fixed by #148, but I didn't phrase it right for that to happen automatically when it was merged.  Closing.

PR #148, opened 2020-02-18
dockerclient: detect ENOENT when copying

nalind:
Detect cases where, when filtering the output of `archiveFromFile()` or `archiveFromContainer()` to net only sources which match a source path, we fail to match anything.  When that happens, return an `os.ErrNotExist` error from the archive's `Close()` method.  The `archiveFromContainer()` function now returns a `ReadCloser` where it previously returned only a `Reader`.

This aims to fix https://bugzilla.redhat.com/show_bug.cgi?id=1794768.

### DIFF
--- a/dockerclient/archive.go
+++ b/dockerclient/archive.go
@@ -205,17 +205,30 @@ func archiveFromFile(file string, src, dst string, excludes []string, check Dire
 	}
 
 	r, err := transformArchive(f, true, mapper.Filter)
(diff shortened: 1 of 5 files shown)


nalind:
Note for reviewers: the dockerclient package isn't used in buildah; we have https://github.com/containers/buildah/pull/2111 to check for this there.

nalind:
This should fix #137.

PR #149, opened 2020-03-12
Process env substiturion in ADD, COPY commands

rhatdan:
In a Containerfile Commands like:

COPY --chown=${NB_USER}:users stuff /stuff

The ${NB_USER} needs to be substituted from the environment.

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -19,9 +19,9 @@ import (
 
 	docker "github.com/fsouza/go-dockerclient"
 
+	"github.com/containerd/containerd/platforms"
 	"github.com/openshift/imagebuilder/signal"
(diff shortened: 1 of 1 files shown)


rhatdan:
This should fix https://github.com/containers/buildah/issues/1806

rhatdan:
@TomSweeneyRedHat @nalind PTAL

rhatdan:
@bparees PTAL

bparees:
can we get a test?

rhatdan:
I was doing the test in buildah, not here.

bparees:
> I was doing the test in buildah, not here.

ok

/lgtm

PR #151, opened 2020-03-13
Treat args the same way `docker build` does

carbonin:
Before this change a Dockerfile like this:

```
FROM alpine
ARG SECRET
RUN echo $SECRET

FROM alpine
RUN echo "$SECRET" > test_file
```

would expose the arg "SECRET" in every layer of the build.

Now the ARG is only accessible in the layer it is defined.
This gives us parity with the Docker behavior described in
https://docs.docker.com/engine/reference/builder/#scope

"An ARG instruction goes out of scope at the end of the build
stage where it was defined. To use an arg in multiple stages,
each stage must include the ARG instruction."

This was originally reported as https://github.com/containers/buildah/issues/2210
cc @rhatdan 

I wasn't sure how to run the tests locally so I hope this one works... :laughing: 

--------------------------------------------------------------------------------------------------
EDIT:

This PR now attempts to handle all the gaps between `imagebuilder`'s ARG handling and `docker build`'s, one of which was the issue described above and in https://github.com/containers/buildah/issues/2210

I have tried to create test cases or edit existing ones to cover all of the cases, but I'll also list them here.

#### An ARG should only be available after its ARG command in the current stage

This is the original problem, described above. As well as the issue raised by @bparees in https://github.com/openshift/imagebuilder/pull/151#issuecomment-598824416

#### A later ARG default value should override an earlier one in the same stage
```Dockerfile
FROM alpine
ARG FOO=foo
ARG FOO=bar
RUN echo "$FOO"
```
The above Dockerfile should print "bar". Before this patch, the behavior was the opposite, an arg was not changed once set (this old behavior effectively achieved the user-provided case, but the combination of the two necessitated a new Builder struct field for user args).

#### User provided args should override all args default values
```Dockerfile
ARG TAG=latest
FROM centos:$TAG
ARG TAG
RUN echo "$TAG"

FROM alpine
ARG TAG="1.2.3"
RUN echo "$TAG"
```

Building the above Docerfile with `--build-arg TAG=7` should build the first stage from `centos:7` and print "7" in both stages.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -209,12 +209,8 @@ func NewStages(node *parser.Node, b *Builder) (Stages, error) {
 		stages = append(stages, Stage{
 			Position: i,
 			Name:     name,
-			Builder: &Builder{
-				Args:        b.Args,
-				AllowedArgs: b.AllowedArgs,
-				Env:         b.Env,
-			},
-			Node: root,
+			Builder:  b.builderForStage(),
+			Node:     root,
 		})
 	}
 	return stages, nil
@@ -235,17 +231,30 @@ func (b *Builder) extractHeadingArgsFromNode(node *parser.Node) error {
 		}
 	}
 
+	// Set children equal to everything except the leading ARG nodes
+	node.Children = children
+
+	// Use a separate builder to evaluate the heading args
+	tempBuilder := NewBuilder(b.UserArgs)
+
+	// Evaluate all the heading arg commands
 	for _, c := range args {
-		step := b.Step()
+		step := tempBuilder.Step()
 		if err := step.Resolve(c); err != nil {
 			return err
 		}
-		if err := b.Run(step, NoopExecutor, false); err != nil {
+		if err := tempBuilder.Run(step, NoopExecutor, false); err != nil {
 			return err
 		}
 	}
 
-	node.Children = children
+	// Add all of the defined heading args to the original builder's HeadingArgs map
+	for k, v := range tempBuilder.Args {
+		if _, ok := tempBuilder.AllowedArgs[k]; ok {
+			b.HeadingArgs[k] = v
+		}
+	}
+
 	return nil
 }
 
@@ -264,13 +273,23 @@ func extractNameFromNode(node *parser.Node) (string, bool) {
 	return n.Next.Value, true
 }
 
+func (b *Builder) builderForStage() *Builder {
+	stageBuilder := NewBuilder(b.UserArgs)
+	for k, v := range b.HeadingArgs {
+		stageBuilder.HeadingArgs[k] = v
+	}
+	return stageBuilder
+}
+
 type Builder struct {
 	RunConfig docker.Config
 
-	Env    []string
-	Args   map[string]string
-	CmdSet bool
-	Author string
+	Env         []string
+	Args        map[string]string
+	HeadingArgs map[string]string
+	UserArgs    map[string]string
+	CmdSet      bool
+	Author      string
 
 	AllowedArgs map[string]bool
 	Volumes     VolumeSet
@@ -288,12 +307,16 @@ func NewBuilder(args map[string]string) *Builder {
 	for k, v := range builtinAllowedBuildArgs {
 		allowed[k] = v
 	}
-	provided := make(map[string]string)
+	userArgs := make(map[string]string)
+	initialArgs := make(map[string]string)
 	for k, v := range args {
-		provided[k] = v
+		userArgs[k] = v
+		initialArgs[k] = v
 	}
 	return &Builder{
-		Args:        provided,
+		Args:        initialArgs,
+		UserArgs:    userArgs,
+		HeadingArgs: make(map[string]string),
 		AllowedArgs: allowed,
 	}
 }
--- a/builder_test.go
+++ b/builder_test.go
@@ -7,11 +7,14 @@ import (
 	"io/ioutil"
 	"os"
 	"reflect"
+	"regexp"
 	"strings"
 	"testing"
 	"time"
 
 	docker "github.com/fsouza/go-dockerclient"
+
+	"github.com/openshift/imagebuilder/dockerfile/parser"
 )
 
 func TestVolumeSet(t *testing.T) {
@@ -159,6 +162,19 @@ func TestMultiStageParseHeadingArg(t *testing.T) {
 	if len(stages) != 3 {
 		t.Fatalf("expected 3 stages, got %d", len(stages))
 	}
+
+	fromImages := []string{"golang:1.9", "busybox:latest", "golang:1.9"}
+	for i, stage := range stages {
+		from, err := stage.Builder.From(stage.Node)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if expected := fromImages[i]; from != expected {
+			t.Fatalf("expected %s, got %s", expected, from)
+		}
+	}
+
 	t.Logf("stages: %#v", stages)
 }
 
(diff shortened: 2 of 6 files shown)


rhatdan:
@TomSweeneyRedHat PTAL

rhatdan:
@bparees PTAL

bparees:
will this handle:

```
ARG FOO=centos
FROM $FOO
RUN echo hello $FOO
```

which in docker at least, does substitute in the FROM line, but does not do anything in the RUN:

```
$ docker build --pull --no-cache .
Sending build context to Docker daemon   5.12kB
Step 1/4 : ARG FOO=centos
Step 2/4 : FROM $FOO
latest: Pulling from library/centos
Digest: sha256:fe8d824220415eed5477b63addf40fb06c3b049404242b31982106ac204f6700
Status: Image is up to date for centos:latest
 ---> 470671670cac
Step 3/4 : RUN echo hello $FOO
 ---> Running in 695dfdf3b572
hello
Removing intermediate container 695dfdf3b572
 ---> 09813fb15eec
Step 4/4 : RUN exit 1
 ---> Running in 23dcc64b005b
The command '/bin/sh -c exit 1' returned a non-zero code: 1
```

carbonin:
@bparees actually it looks like that's an error with this change.

master branch:
```
$ ../imagebuilder .
--> Image centos was not found, pulling ...
--> FROM centos as 0
--> RUN echo hello $FOO
hello centos
--> Committing changes ...
--> Done
```

This patch:
```
$ ../imagebuilder .
json: cannot unmarshal array into Go value of type docker.Image
```

Will look into that when I get a chance.

carbonin:
@bparees It looks like https://github.com/openshift/imagebuilder/blob/master/builder_test.go#L165 should fail in this case, right?

Sorry for the basic question, but how do I run that test case?

It doesn't seem to run using `make test` as those passed for me before submitting this.
`make test-conformance` fails for me with this change and on `master` with this error:
```
$ make test-conformance
go test -v -tags conformance -timeout 10m ./dockerclient
# github.com/openshift/imagebuilder/dockerclient
dockerclient/archive.go:227:11: Errorf format %w has unknown verb w
FAIL	github.com/openshift/imagebuilder/dockerclient [build failed]
make: *** [Makefile:10: test-conformance] Error 2
```

bparees:
> should fail in this case, right?

i don't think that would fail, it's just going to resolve to `FROM busybox:`    (empty tag).

which means it's not a great test.

bparees:
oh sorry, i see what you mean.  i do not know, sorry.  this really isn't a repo i deal w/ normally

carbonin:
@TomSweeneyRedHat ?

TomSweeneyRedHat:
Sorry @carbonin I've been buried in a gazillion things recently.  Hope to get a good look this weekend.

builder.go
```diff
@@ -264,6 +264,16 @@ func extractNameFromNode(node *parser.Node) (string, bool) {
 	return n.Next.Value, true
 }
 
+func argsForStage(stageRoot *parser.Node, allArgs map[string]string) map[string]string {
```
TomSweeneyRedHat: For history's sake, it would be good to have a comment for this function talking about what you expect it to do.

TomSweeneyRedHat:
I too have not been in this project for several months, but I know @nalin has been quite a bit.  So if he says anything different than I, trust his word.

First off, something is off with archive.go.  I've just fixed that in #152, however now the conformance tests are failing in a number of places with that fix in play.  I'm guessing that the conformance tests are not being regularly run.

To run a particular test in builder_test.go, you need to do (-v (verbose) is optional, but very helpful):

`go test -v -run TestByTarget`   where `TestByTarget` is a function name in bulider_test.go.
or
`go test -v -tags archive ./dockerclient` #archive tests
or
`go test -v -tags conformance ./dockerclient`  #conformance tests

But those too are also not passing.  However, your new test seems to be working:

```
# go test -v -run TestMultiStageArgScope
=== RUN   TestMultiStageArgScope
--- PASS: TestMultiStageArgScope (0.00s)
PASS
ok  	github.com/openshift/imagebuilder	0.004s
```
Once you chase down the problem above, it would be good to get @nalin to eyeball this.

carbonin:
Okay I added a commit to alter an existing test case to deal with the situation from [@bparees's comment](https://github.com/openshift/imagebuilder/pull/151#issuecomment-598824416)

But while investigating this behavior and how to deal with it I found https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact#understand-how-arg-and-from-interact which complicates things a bit.

What behavior should we be striving for here? Are we trying to emulate docker exactly? There seems to be considerable confusion around this particular command combination: https://github.com/moby/moby/issues/34129

Right now we're treating the "heading" args as global and applying them to all the following commands, but Docker's handling is only applying the heading args to `FROM` commands unless they are re-"defined" in a particular stage (in which case the default, if any, is inherited from the first definition).

Is it safe to assume we don't want to tackle all of that in this patch? With that assumption, how far should I take this one? Just fix the test in the new commit I added?

I think restricting the scope of the heading args would be a rather breaking change for users of `buildah`, but maybe just as breaking as scoping args to a particular stage.

carbonin:
Oh and the edited test now fails like this with my changes:

```
$ go test -v -run TestMultiStageParseHeadingArg
=== RUN   TestMultiStageParseHeadingArg
--- FAIL: TestMultiStageParseHeadingArg (0.00s)
    builder_test.go:170: expected golang:1.9, got :
FAIL
exit status 1
FAIL	github.com/openshift/imagebuilder	0.003s
```

bparees:
> What behavior should we be striving for here? Are we trying to emulate docker exactly? 

for imagebuilder itself i don't have a strong opinion, but for buildah as a consumer of imagebuilder, i think buildah should be emulating docker including all its...idiosyncrasies, since it is routed as a drop-in/aliasable replacement for "docker build"


it may be worth buildah having a "--sane" mode which drops those compatbility behaviors in deference to preferred behavior.  (thinking about the recent COPY behavior discoveries we have made)

TomSweeneyRedHat:
I'd just tweak @bparees expectation comment slightly.  `buildah bud --layers` should be equivalent to `docker build`.  We advertise that `podman build` is equivalent to `docker build`.  Podman calls `buildah bud` under the covers and when it does so it adds the `--layers` option.

carbonin:
Okay, so let me ask a more specific question. How do we want `NewStages` to behave?

Each stage returned by that function has a Builder with args. Should every builder have access to the args defined before the first `FROM` command when returned from `NewStages`? I assume we also want to define explicitly how something really awful like this would behave, right?

```Dockerfile
ARG FOO=centos
ARG BAZ=things

FROM $FOO
RUN echo hello $FOO

FROM alpine
RUN echo hello $FOO
ARG FOO=alpine

FROM $FOO
ARG BAZ
RUN echo hello $FOO
RUN echo hello $BAZ
```

For what it's worth, right now it's like this:
```
--> FROM centos as 0
--> RUN echo hello $FOO
hello centos
--> FROM alpine as 1
--> RUN echo hello $FOO
hello centos
--> ARG FOO=alpine
--> FROM centos as 2
--> ARG BAZ
--> RUN echo hello $FOO
hello centos
--> RUN echo hello $BAZ
hello things
--> Committing changes ...
--> Done
```

Buildah can post-process those stages to get what we want there. But what, for this project, do we want the "correct" behavior to be?

carbonin:
Note that in the previous example the first `ARG FOO` default value is not overridden by the next default even though nothing was passed in the command line

carbonin:
Okay, this now tries to tackle all of the gaps I could find between `imagebuilder` and `docker` relating to `ARG` handling.

I found one more bug that I'm still trying to work out, but would appreciate a look at what I have here as this is getting quite involved.

rhatdan:
@carbonin Could you open a WIP vendor of this in Buildah to make sure it passes the Buildah CI tests.

rhatdan:
The code looks good to me, and I like that there are lots of tests.  The issue is probably getting people more in the know of the deep level of how this is supposed to work.

@kolyshkin PTAL

carbonin:
I've went full circle here and removed the method I added in the first commit :sweat_smile: 

I can squash these if that's preferred ...

carbonin:
> Could you open a WIP vendor of this in Buildah to make sure it passes the Buildah CI tests.

@rhatdan added a commit in https://github.com/containers/buildah/pull/2220 to include my branch as the imagebuilder source

rhatdan:
@carbonin Thanks, it will give me better confidence on this if it can pass Buildah and imagebuilder tests.

carbonin:
@rhatdan Seeing as how this is passing tests here and the only tests failing in buildah with this branch are the ones I added to expose this problem, can we move forward with this patch?

Is there someone else I should contact for review?

rhatdan:
lgtm
@nalind @TomSweeneyRedHat @bparees PTAL

nalind:
Patches look reasonable, and the behavior is consistent with what I can observe with the moby-engine-18.09.7-5.ce.git2d0083d.fc30 package on Fedora 30.  LGTM.

PR #153, opened 2020-03-16
archiveFromContainer(): return possibly-wrapped os.IsNotError

nalind:
When a source specified as a glob can't be found, return an error that can be identified by as an `os.ErrIsNotExist`, wrapped or otherwise, even if we have to drop the glob from the error string to ensure that for versions of Go that don't wrap errors.

Unit tests for the `dockerclient` package were failing with Go 1.12 without this patch.  See the [1.13 release note](https://golang.org/doc/go1.13#error_wrapping) on the use of the `%w` format specifier.

### DIFF
--- a/dockerclient/archive.go
+++ b/dockerclient/archive.go
@@ -224,7 +224,7 @@ func archiveFromContainer(in io.Reader, src, dst string, excludes []string, chec
 	r, err := transformArchive(in, false, mapper.Filter)
 	rc := readCloser{Reader: r, Closer: newCloser(func() error {
 		if !mapper.foundItems {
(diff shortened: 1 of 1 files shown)


rhatdan:
It is probably time we drop go 1.12.1...
But LGTM

rhatdan:
@TomSweeneyRedHat @bparees PTAL

TomSweeneyRedHat:
Thanks @nalind.  I looked for `%w` over the weekend, but had didn't run across it via google.  This is a neat solution, thx for the pointers.

TomSweeneyRedHat:
and forgot.
LGTM

PR #155, opened 2020-04-28
Need to pass down user account when creating the workdir

rhatdan:
Currently the Workingdir does not have the correct protection
if your Containerfile sets the USER field before the WORKDIR
field, the WORKDIR should be owned by the user.

This needs to be passed back to the stage.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -40,7 +40,7 @@ type Run struct {
 
 type Executor interface {
 	Preserve(path string) error
-	EnsureContainerPath(path string) error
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Changes LGTM,but tests don't look terribly happy.

rhatdan:
After testing the fix with Docker, the assumptions are wrong, and the WORKDIR does not get set with ownership of the current USER.  So this PR is not needed.

Issue #160, opened 2020-05-19
TestDispatchArgDefaultBuiltins is failing on arm/v8


siretart:
I've packaged this project for debian, where we build it on a large number of [different platforms](https://buildd.debian.org/status/package.php?p=golang-github-openshift-imagebuilder&suite=sid). It turns out that the update from verison 1.0 to 1.1.4 broke the test TestDispatchArgDefaultBuiltins on all Arm platforms (links to buildlogs for: [arm64](https://buildd.debian.org/status/fetch.php?pkg=golang-github-openshift-imagebuilder&arch=arm64&ver=1.1.4-1&stamp=1588027968&raw=0), [armel](https://buildd.debian.org/status/fetch.php?pkg=golang-github-openshift-imagebuilder&arch=armel&ver=1.1.4-1&stamp=1588028270&raw=0), and [armhf](https://buildd.debian.org/status/fetch.php?pkg=golang-github-openshift-imagebuilder&arch=armhf&ver=1.1.4-1&stamp=1588028207&raw=0))

The symptom looks always like this.

```
=== RUN   TestDispatchArgDefaultBuiltins
    TestDispatchArgDefaultBuiltins: dispatchers_test.go:31: Expected [BUILDARCH=arm64 TARGETPLATFORM=linux/arm64], got [BUILDARCH=arm64 TARGETPLATFORM=linux/arm64/v8]
--- FAIL: TestDispatchArgDefaultBuiltins (0.00s)
```

This test code is here: https://github.com/openshift/imagebuilder/blob/e3888005bf9062f32a83f209195572cc113db57e/dispatchers_test.go#L13-L33

It was introduced in https://github.com/openshift/imagebuilder/commit/54d2f38195e4c9deb084fbc230eabc58f9b6bf06

If I had to guess, I think that the following condition is only entered in ARM platforms and that's not covered by the test: https://github.com/openshift/imagebuilder/blob/e3888005bf9062f32a83f209195572cc113db57e/dispatchers.go#L586-L588

@grooverdan any chance you could have a look and form an opinion on what's the right approach to fix this?

PR #161, opened 2020-05-19
TestDispatchArgDefaultBuiltins: Unbreak on arm

siretart:
On ARM, the variant is non-empty and reflected in the TARGETPLATFORM default string

Fixes: #160

### DIFF
--- a/dispatchers_test.go
+++ b/dispatchers_test.go
@@ -23,7 +23,7 @@ func TestDispatchArgDefaultBuiltins(t *testing.T) {
 	localspec := platforms.DefaultSpec()
 	expectedArgs := []string{
 		"BUILDARCH=" + localspec.Architecture,
-		"TARGETPLATFORM=" + localspec.OS + "/" + localspec.Architecture,
+		"TARGETPLATFORM=" + platforms.DefaultString(),
 	}
 	got := mybuilder.Arguments()
 	sort.Strings(got)


dispatchers_test.go
```diff
@@ -23,7 +23,7 @@ func TestDispatchArgDefaultBuiltins(t *testing.T) {
 	localspec := platforms.DefaultSpec()
 	expectedArgs := []string{
 		"BUILDARCH=" + localspec.Architecture,
-		"TARGETPLATFORM=" + localspec.OS + "/" + localspec.Architecture,
+		"TARGETPLATFORM=" + platforms.DefaultString(),
```
TomSweeneyRedHat, line 26: Should the equivalent in [dispatchers.go](https://github.com/openshift/imagebuilder/blob/8de717452756bdf0348f0393d1205b095d80bdb8/dispatchers.go#L35) be changed too?  IDK for sure, but thought I'd ask.

siretart: Possibly, but then https://github.com/openshift/imagebuilder/blob/8de717452756bdf0348f0393d1205b095d80bdb8/dispatchers.go#L45-L50 can probably go away as well. Not sure what other consequences this would have on the testsuite and elswhere.

rhatdan:
@siretart Are you still interested in this PR?  Are you going to make the suggested changes?

PR #159, opened 2020-05-15
Allow Args to be used in Variable Evaluation

TomSweeneyRedHat:
If the Dockerfile has a `ARG CHOWN_VAL=6173:6173`, then
a `COPY --chown=${CHOWN_VAL} srcFile destFile` will fail.
However if the Dockerfile has a `ENV CHOWN_VAL=6173:6173`
then it succeeds.

The dispatchCopy() and add() functions were only going through the
list of Environment variables to resolve a Variable instead
of the list of Args and Environment variables.  

Now add the definitions from the ARG statements to the ones
from the ENV statements and use both when resolving 
variables.

At some point I'd like to look into converting the b.Env to
a Map from a slice like b.Args, but didn't want to run this
down now.

Fixes: https://github.com/containers/buildah/issues/2192
       and probably https://github.com/containers/buildah/issues/2345

Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

PR #163, opened 2020-05-29
Handle ARG and ENV with duplicate name on RUN

TomSweeneyRedHat:
If a Dockerfile had an ARG and an ENV with the
same name, things become very unpredictable in
the RUN statement.  This change moves the
code created for handling this same issue in
COPY to the the internals.go file so it
can be used by dispatchers.go and builder.go.

Use the makeUserArgs() function to create a list
of variables, giving precedence to any variables
defined by the ENV statement(s).

This issue was first noted in #159 and once vendored,
this will fix: https://github.com/containers/buildah/issues/2372

### DIFF
--- a/builder_test.go
+++ b/builder_test.go
@@ -775,3 +775,47 @@ func TestBuilder(t *testing.T) {
 		})
 	}
 }
+
+func TestRunWithEnvArgConflict(t *testing.T) {
+	f, err := os.Open("dockerclient/testdata/Dockerfile.envargconflict")
+	if err != nil {
+		t.Fatal(err)
+	}
+	node, err := ParseDockerfile(f)
+	if err != nil {
+		t.Fatal(err)
+	}
+	b := NewBuilder(nil)
+	from, err := b.From(node)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
@nalind @vrothberg @QiWang19 PTAL

builder_test.go
```diff
@@ -775,3 +775,43 @@ func TestBuilder(t *testing.T) {
...
+		if err := step.Resolve(child); err != nil {
+			t.Fatal(err)
+		}
+		if err := b.Run(step, LogExecutor, false); err != nil {
+			t.Fatal(err)
+		}
+	}
+	configString := fmt.Sprintf("%v", b.Config())
+	expectedValue := "USER_NAME=my_user_env"
+	if !strings.Contains(configString, expectedValue) {
```
QiWang19, line 807: Maybe add a test to check `my_user_arg` value does not contain?

TomSweeneyRedHat: Good thought @QiWang19 .  Added and pushed.

QiWang19:
LGTM

PR #164, opened 2020-06-01
Adjust argument slice size in builder

TomSweeneyRedHat:
A recent update to the argument handling didn't increase
the argument slice size to take into account the values
from b.Args() that were now being added.  The created
unpredictable results.


builder.go
```diff
@@ -332,7 +332,7 @@ func ParseFile(path string) (*parser.Node, error) {
 
 // Step creates a new step from the current state.
 func (b *Builder) Step() *Step {
-	dst := make([]string, len(b.Env)+len(b.RunConfig.Env))
+	dst := make([]string, len(b.Env)+len(b.RunConfig.Env)+len(b.Arguments()))
 	copy(dst, makeUserArgs(b))
 	dst = append(dst, b.RunConfig.Env...)
```
nalind, line 337: Some of the values from `b.Args` won't be used if they conflict with environment values, so the size count being used here for allocating `dst` can be off, and if it is, it can leave a gap with empty strings in the final slice.

rhatdan:
LGTM

PR #167, opened 2020-06-26
Handle args with multiple equal signs

TomSweeneyRedHat:
If an ARG in a Dockerfile looks like this:

ARG multivalarg="a=1 b=2 c=3 d=4"

the processing would drop everything after the
second equal sign leaving multivalarg="a".

Addresses: https://github.com/containers/buildah/issues/2424


QiWang19:
LGTM

TomSweeneyRedHat:
@rhatdan PTAL

PR #169, opened 2020-07-13
Deduplicate environment variables

nalind:
When building lists of environment variables, or environment variables combined with build arguments, always deduplicate sets of values.  This should fix https://bugzilla.redhat.com/show_bug.cgi?id=1843727.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -332,20 +332,10 @@ func ParseFile(path string) (*parser.Node, error) {
 
 // Step creates a new step from the current state.
 func (b *Builder) Step() *Step {
-	argsMap := make(map[string]string)
-	for _, argsVal := range b.Arguments() {
-		val := strings.SplitN(argsVal, "=", 2)
-		if len(val) > 1 {
-			argsMap[val[0]] = val[1]
-		}
-	}
-
-	userArgs := makeUserArgs(b.Env, argsMap) 
-	dst := make([]string, len(userArgs)+len(b.RunConfig.Env))
-	copy(dst, userArgs)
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat @ashley-cui PTAL

builder.go
```diff
@@ -332,20 +332,7 @@ func ParseFile(path string) (*parser.Node, error) {
...
-		}
-	}
-
-	userArgs := makeUserArgs(b.Env, argsMap) 
-	dst := make([]string, len(userArgs)+len(b.RunConfig.Env))
-	copy(dst, userArgs)
-	dst = append(dst, b.RunConfig.Env...)
-
-	return &Step{Env: dst}
+	return &Step{Env: mergeEnv(b.Arguments(), mergeEnv(b.Env, b.RunConfig.Env))}
```
TomSweeneyRedHat, line 335: This is really slick.  Could we add a code comment here talking about this ordering being important so that the values in b.Arguments() won't be added to the list if the key was already created via values found in the ENV fields?

TomSweeneyRedHat:
One question/suggestion, otherwise LGTM.

builder.go
nalind: Added comment `Include build arguments in the table of variables that we'll use in Resolve(), but override them with values from the actual environment in case there's any conflict.`  Happy to tweak the wording if that's omitting something.

TomSweeneyRedHat: That looks good to me.  I was just concerned that a year from now I'd forget the importance of that ordering.

PR #170, opened 2020-07-14
Update parsing of quoted strings

nalind:
When parsing a single-quoted token, if we reach an EOF before we reach the closing quote, flag it as an error.
When parsing a double-quoted token, accept escaped escape characters ("\\"), and if we reach an EOF before we reach the closing quote, flag it as an error.
This should bring our parsing in line with docker build `17.06` and later.


rhatdan:
LGTM
@TomSweeneyRedHat PTAL

TomSweeneyRedHat:
@nalind I still have a minor, but unanswered question on this one....

PR #174, opened 2020-08-03
Fix accidentally discarding the environment for ENV

### LINKED ISSUE #173: Regression in ENV expansion
There has been a recent regression in expansion of ENV since 1.1.6.  Test case:
``` Dockerfile
FROM registry.redhat.io/ubi8

ENV FOO_DIR=/opt/foo
ENV PATH=$FOO_DIR/bin:$PATH

RUN ls -l
```
With 1.1.6 or earlier, this works as expected.  With git master it falis:
```
--> RUN ls -l
/bin/sh: ls: command not found
```
My guess is https://github.com/openshift/imagebuilder/pull/169 may be the culprit but I haven't bisected it.

(The real-life case was https://github.com/kube-reporting/hadoop fwiw.)

nalind:
When processing `ENV` instructions, we were mistakenly replacing the contents of `Env` with the contents of `RunConfig.Env`.

Fixes #173.

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -85,7 +85,7 @@ func env(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 		// value ==> args[j+1]
 		newVar := []string{args[j] + "=" + args[j+1]}
 		b.RunConfig.Env = mergeEnv(b.RunConfig.Env, newVar)
-		b.Env = mergeEnv(b.RunConfig.Env, newVar)
+		b.Env = mergeEnv(b.Env, newVar)
 		j++
 	}
 


dispatchers.go
```diff
@@ -85,7 +85,7 @@ func env(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 		// value ==> args[j+1]
 		newVar := []string{args[j] + "=" + args[j+1]}
 		b.RunConfig.Env = mergeEnv(b.RunConfig.Env, newVar)
-		b.Env = mergeEnv(b.RunConfig.Env, newVar)
+		b.Env = mergeEnv(b.Env, newVar)
```
TomSweeneyRedHat, line 88: Dope slap here.  Missed that in the earlier review.

PR #176, opened 2020-09-16
ADD: extract local archives

### LINKED ISSUE #99: ADD doesn't expand archives
There's this fun Docker behavior where ADD (but not COPY!) will expand a tarball instead of just copying it into place as-is. 

[If \<src\> is a local tar archive in a recognized compression format (identity, gzip, bzip2 or xz) then it is unpacked as a directory. Resources from remote URLs are not decompressed. ... Whether a file is identified as a recognized compression format or not is done solely based on the contents of the file, not the name of the file. ](https://docs.docker.com/engine/reference/builder/#add)

imagebuilder doesn't do that. This would be useful to expand source archives from lookaside in OSBS. You can work around it with `RUN tar fx` or whatever but it's one more difference to have to account for.

```
🐚 cd $(mktemp -d)
[tmp.aqbI4PmaDS] 🐚 touch foo 
[tmp.aqbI4PmaDS] 🐚 tar fc foo.tar foo
[tmp.aqbI4PmaDS] 🐚 cat <<DF > Dockerfile
> FROM fedora:28
> WORKDIR /tmp
> ADD foo.tar .
> RUN ls | grep foo
> DF
[tmp.aqbI4PmaDS] 🐚 imagebuilder .
--> FROM fedora:28 as 0
--> WORKDIR /tmp
--> ADD foo.tar .
--> RUN ls | grep foo
foo.tar
--> Committing changes ...
--> Done
[tmp.aqbI4PmaDS] 🐚 
[tmp.aqbI4PmaDS] 🐚 
[tmp.aqbI4PmaDS] 🐚 docker build .
Sending build context to Docker daemon 13.31 kB
Step 1/4 : FROM fedora:28
 ---> 75aeb7f897fd
Step 2/4 : WORKDIR /tmp
 ---> Using cache
 ---> 41c3743a8378
Step 3/4 : ADD foo.tar .
 ---> 329f459cf92a
Removing intermediate container 80b097ebbd16
Step 4/4 : RUN ls | grep foo
 ---> Running in 34e030016475
foo
 ---> c5ae75b6e86b
Removing intermediate container 34e030016475
Successfully built c5ae75b6e86b
```

nalind:
When ADDing files from the build context that look like archives, the expected behavior is that we'll extract their contents in the destination location.  Fixes #99.


dockerclient/testdata/add/archived.txt
```diff
@@ -0,0 +1 @@
+help, i've been archived
```
TomSweeneyRedHat, line 1: I've heard being archived is very painful....
;^)

dockerclient/client.go
```diff
@@ -923,6 +923,28 @@ func (e *ClientExecutor) Archive(fromFS bool, src, dst string, allowDownload boo
...
+		}
+		klog.V(5).Infof("Extracting %s -> %s from local archive", srcFullPath, dst)
+		transform := func(h *tar.Header, r io.Reader) (data []byte, update bool, skip bool, err error) {
+			h.Name = filepath.Join(dst, h.Name)
+			if h.Typeflag == tar.TypeLink {
+				h.Linkname = filepath.Join(dst, h.Linkname)
+			}
+			return nil, false, false, nil
+		}
+		filtered, err := transformArchive(f, true, transform)
```
TomSweeneyRedHat, line 941: might be end of day brain, but where does `transformArchive` live?

rhatdan:
LGTM

dockerclient/client.go
nalind: It's in [archive.go](https://github.com/openshift/imagebuilder/blob/master/dockerclient/archive.go#L234).  This part of imagebuilder is the direct inspiration for similar functionality [in buildah](https://github.com/containers/buildah/blob/master/digester.go#L90).

nalind:
Realized I'd changed the behavior for COPY, too.  Added a check on `allowDownload`, and added a conformance test that looks just like the new one we were already adding, but using COPY instead of ADD, so that we verify that we don't extract archives for COPY.

TomSweeneyRedHat:
LGTM

PR #177, opened 2020-10-12
Support comments in .dockerignore

nalind:
When reading the contents of .dockerignore, filter out empty lines and lines that have a '#' as their first character.  This will address https://github.com/containers/buildah/issues/2686, which also affects us.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -563,14 +563,21 @@ var builtinAllowedBuildArgs = map[string]bool{
 }
 
 // ParseDockerIgnore returns a list of the excludes in the .dockerignore file.
-// extracted from fsouza/go-dockerclient.
+// extracted from fsouza/go-dockerclient and modified to drop comments and
+// empty lines.
 func ParseDockerignore(root string) ([]string, error) {
 	var excludes []string
 	ignore, err := ioutil.ReadFile(filepath.Join(root, ".dockerignore"))
 	if err != nil && !os.IsNotExist(err) {
 		return excludes, fmt.Errorf("error reading .dockerignore: '%s'", err)
 	}
-	return strings.Split(string(ignore), "\n"), nil
+	for _, e := range strings.Split(string(ignore), "\n") {
+		if len(e) == 0 || e[0] == '#' {
+			continue
+		}
+		excludes = append(excludes, e)
+	}
+	return excludes, nil
 }
 
 // ExportEnv creates an export statement for a shell that contains all of the
--- a/builder_test.go
+++ b/builder_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"strings"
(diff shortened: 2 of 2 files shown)


builder_test.go
```diff
@@ -856,3 +857,49 @@ func TestRunWithMultiArg(t *testing.T) {
...
+		},
+		{
+			input:  []string{"", "first", "second", "", " #third", "#invalid pattern which shouldn't matter ("},
+			result: []string{"first", "second", " #third"},
+		},
+		{
+			input:  []string{"", "first", "second", "", "#third", ""},
+			result: []string{"first", "second"},
+		},
+	}
```
TomSweeneyRedHat, line 883: Can you create an input that has a leading `#`?

builder.go
```diff
@@ -563,14 +563,21 @@ var builtinAllowedBuildArgs = map[string]bool{
...
+// empty lines.
 func ParseDockerignore(root string) ([]string, error) {
 	var excludes []string
 	ignore, err := ioutil.ReadFile(filepath.Join(root, ".dockerignore"))
 	if err != nil && !os.IsNotExist(err) {
 		return excludes, fmt.Errorf("error reading .dockerignore: '%s'", err)
 	}
-	return strings.Split(string(ignore), "\n"), nil
+	for _, e := range strings.Split(string(ignore), "\n") {
+		if len(e) == 0 || e[0] == '#' {
```
TomSweeneyRedHat, line 575: should this be
```suggestion
		if len(e) == 0 || strings.HasPrefix(e[0] == '#') {
```
or do we process commented lines if they're not empty?

builder_test.go
nalind: Do you mean on the first line?  The fifth line in this file has one.

builder.go
nalind: This should be skipping all empty or commented lines (well, commented in the first column, per docs).  Are you seeing it missing a case that it should be catching?

builder_test.go
TomSweeneyRedHat: Yes, I think it would be good to have one on the first line too.

builder.go
TomSweeneyRedHat: I may have the logic reveresed here, but here's what I was looking at.  Given a string `# Test`, I expected that line to be ignored (hit the continue at line 576).   When I ran a small piece of go code,  the if statement in line 575 did not equate to 'True` for that line as `e[0]` contained `# Test` and not `#`.  I was thinking it should fire for that line, or have I been playing too much reversi lately?

nalind: `strings.Split()` should return a slice of strings, so if we're iterating through that, `e` will be a string, and its elements will be runes.

builder_test.go
nalind: Sure, adding it.

rhatdan: One of the tests should have the `#` in the middle of a line (word)
```
                         input:  []string{"", "fi#rst", "#se#cond", "", " #third", "#invalid pattern which shouldn't matter ("},
			result: []string{"fi#rst", " #third"},
```

rhatdan:
LGTM
Other then perhaps adding one more test.

builder_test.go
nalind: Sure.  Added.

builder.go
TomSweeneyRedHat: @nalind Thanks, that was my bad, I was working on an array of strings in my tests.  Reworked my test and it fires as you intended.

TomSweeneyRedHat (commented):
LGTM

PR #178, opened 2020-10-13
[release-3.11] support comments in .dockerignore

nalind:
When reading the contents of .dockerignore, filter out empty lines and lines that have a '#' as their first character, from #177.


nalind:
Hmm, I guess we'll hold off on this one until we confirm that it's affecting people.

PR #179, opened 2020-10-13
ADD: add another conformance test

nalind:
Make sure that when we're extracting archives that are ADDed, we don't extract archives under a directory that we're ADDing, so that we won't need a fix like https://github.com/containers/buildah/pull/2667 in the future.


TomSweeneyRedHat:
LGTM
and happy green test buttons

PR #183, opened 2021-02-10
Support .containerignore as well as .dockerignore

rhatdan:
We have been asked in buildah to have containerignore as an alternative
name for dockerignore, similer to Containerfile versus Dockerfile.

This pr will look for .containerfile first and then fail over to
.dockerignore.

We also want to add a new feature to support
buildah bud -f FILENAME.dockerignore /tmp

This is a new feature of Docker to allow people to create a file
with .dockerignore postfix, instead of just looking for .dockerignore

Ignore begining or trailing '/' in ignore line

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -562,14 +562,16 @@ var builtinAllowedBuildArgs = map[string]bool{
 	"no_proxy":    true,
 }
 
-// ParseDockerIgnore returns a list of the excludes in the .dockerignore file.
+// ParseIgnore returns a list of the excludes in the specified path
+// path should be a file with the .dockerignore format
 // extracted from fsouza/go-dockerclient and modified to drop comments and
 // empty lines.
-func ParseDockerignore(root string) ([]string, error) {
+func ParseIgnore(path string) ([]string, error) {
 	var excludes []string
-	ignore, err := ioutil.ReadFile(filepath.Join(root, ".dockerignore"))
-	if err != nil && !os.IsNotExist(err) {
-		return excludes, fmt.Errorf("error reading .dockerignore: '%s'", err)
+
+	ignore, err := ioutil.ReadFile(path)
+	if err != nil {
+		return excludes, err
 	}
 	for _, e := range strings.Split(string(ignore), "\n") {
 		if len(e) == 0 || e[0] == '#' {
@@ -578,6 +580,18 @@ func ParseDockerignore(root string) ([]string, error) {
 		excludes = append(excludes, e)
 	}
 	return excludes, nil
+}
+
+// ParseDockerIgnore returns a list of the excludes in the .containerignore or .dockerignore file.
+func ParseDockerignore(root string) ([]string, error) {
+	excludes, err := ParseIgnore(filepath.Join(root, ".containerignore"))
+	if err != nil && os.IsNotExist(err) {
+		excludes, err = ParseIgnore(filepath.Join(root, ".dockerignore"))
+	}
+	if err != nil && os.IsNotExist(err) {
+		return excludes, nil
+	}
+	return excludes, err
 }
 
 // ExportEnv creates an export statement for a shell that contains all of the
--- a/builder.go
+++ b/builder.go
@@ -569,15 +569,18 @@ var builtinAllowedBuildArgs = map[string]bool{
 func ParseIgnore(path string) ([]string, error) {
 	var excludes []string
 
-	ignore, err := ioutil.ReadFile(path)
+	ignores, err := ioutil.ReadFile(path)
 	if err != nil {
 		return excludes, err
 	}
-	for _, e := range strings.Split(string(ignore), "\n") {
-		if len(e) == 0 || e[0] == '#' {
+	for _, ignore := range strings.Split(string(ignores), "\n") {
+		if len(ignore) == 0 || ignore[0] == '#' {
 			continue
 		}
-		excludes = append(excludes, e)
+		ignore = strings.Trim(ignore, "/")
+		if len(ignore) > 0 {
+			excludes = append(excludes, ignore)
+		}
 	}
 	return excludes, nil
 }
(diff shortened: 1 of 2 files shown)


rhatdan:
@nalind @TomSweeneyRedHat PTAL

Helps fix: https://github.com/containers/buildah/issues/2944

builder.go
```diff
@@ -580,6 +582,15 @@ func ParseDockerignore(root string) ([]string, error) {
 	return excludes, nil
 }
 
+// ParseDockerIgnore returns a list of the excludes in the .containerfile or .dockerignore file.
+func ParseDockerignore(root string) ([]string, error) {
+	excludes, err := ParseIgnore(filepath.Join(root, ".containerignore"))
+	if err != nil && !os.IsNotExist(err) {
+		excludes, err = ParseIgnore(filepath.Join(root, ".containerignore"))
```
TomSweeneyRedHat, line 589: I think you want this here:
```suggestion
		excludes, err = ParseIgnore(filepath.Join(root, ".dockerrignore"))
```

rhatdan: Oops

builder.go
```diff
@@ -580,6 +582,15 @@ func ParseDockerignore(root string) ([]string, error) {
 	return excludes, nil
 }
 
+// ParseDockerIgnore returns a list of the excludes in the .containerfile or .dockerignore file.
+func ParseDockerignore(root string) ([]string, error) {
+	excludes, err := ParseIgnore(filepath.Join(root, ".containerignore"))
+	if err != nil && !os.IsNotExist(err) {
```
nalind, line 588: This condition looks like it should be `if err != nil && os.IsNotExist(err)`.

builder.go
nalind: `.dockerignore`, not `.dockerrignore`

builder.go
rhatdan: Fixed

builder.go
```diff
@@ -562,14 +562,16 @@ var builtinAllowedBuildArgs = map[string]bool{
...
+// path should be a file with the .dockerignore format
 // extracted from fsouza/go-dockerclient and modified to drop comments and
 // empty lines.
-func ParseDockerignore(root string) ([]string, error) {
+func ParseIgnore(path string) ([]string, error) {
 	var excludes []string
-	ignore, err := ioutil.ReadFile(filepath.Join(root, ".dockerignore"))
+
+	ignore, err := ioutil.ReadFile(path)
 	if err != nil && !os.IsNotExist(err) {
```
nalind, line 573:
```suggestion
	if err != nil {
		if os.IsNotExist(err) {
			return excludes, err
		}
```

builder.go
```diff
@@ -580,6 +582,15 @@ func ParseDockerignore(root string) ([]string, error) {
 	return excludes, nil
 }
 
+// ParseDockerIgnore returns a list of the excludes in the .containerignore or .dockerignore file.
+func ParseDockerignore(root string) ([]string, error) {
+	excludes, err := ParseIgnore(filepath.Join(root, ".containerignore"))
+	if err != nil && os.IsNotExist(err) {
+		excludes, err = ParseIgnore(filepath.Join(root, ".dockerignore"))
+	}
```
nalind, line 590:
```suggestion
	}
	if err != nil && os.IsNotExist(err) {
		return excludes, nil
	}
```

builder.go
TomSweeneyRedHat: double oops!

builder.go
```diff
@@ -562,14 +562,19 @@ var builtinAllowedBuildArgs = map[string]bool{
...
-	ignore, err := ioutil.ReadFile(filepath.Join(root, ".dockerignore"))
-	if err != nil && !os.IsNotExist(err) {
-		return excludes, fmt.Errorf("error reading .dockerignore: '%s'", err)
+
+	ignore, err := ioutil.ReadFile(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return excludes, err
+		}
+		return excludes, err
```
TomSweeneyRedHat, line 577: Maybe it's too late at night, but can't we get rid of the inner if here, or were you planning to do other processing there?

nalind:
This needs to add tests.

builder.go
nalind: Doh, you're right, the goal was to expose `IsNotExist` errors to the caller so that `ParseDockerignore()`'s fallback logic would be triggered.

rhatdan:
@nalind @TomSweeneyRedHat Tests added, PTAL

builder.go
```diff
@@ -562,24 +562,39 @@ var builtinAllowedBuildArgs = map[string]bool{
...
 	}
-	for _, e := range strings.Split(string(ignore), "\n") {
-		if len(e) == 0 || e[0] == '#' {
+	for _, ignore := range strings.Split(string(ignores), "\n") {
+		if len(ignore) == 0 || ignore[0] == '#' {
 			continue
 		}
-		excludes = append(excludes, e)
+		// drop  "/" when begingin or ending an ignore string
```
nalind, line 580: Typo: "begingin"

builder.go
```diff
@@ -562,24 +562,39 @@ var builtinAllowedBuildArgs = map[string]bool{
...
+		excludes = append(excludes, strings.TrimPrefix(strings.TrimSuffix(ignore, "/"), "/"))
```
nalind, line 581: The tests don't exercise this logic.

rhatdan: They do now.

builder.go
```diff
@@ -562,24 +562,51 @@ var builtinAllowedBuildArgs = map[string]bool{
...
+			}
+			ignore = strings.TrimPrefix(ignore, "/")
+		}
+		for {
+			if !strings.HasSuffix(ignore, "/") {
+				break
+			}
+			ignore = strings.TrimSuffix(ignore, "/")
+		}
+		excludes = append(excludes, ignore)
```
nalind, line 593: Maybe use `strings.Trim()` here, and guard against the end-result of the trimming being an empty string.

rhatdan: Nice, I did not know this was available.

TomSweeneyRedHat:
LGTM
would like a @nalind head nod

nalind:
Conformance tests pass.
/lgtm

PR #184, opened 2021-02-16
Add chmod handling to COPY/ADD

jmguzik:
In order to properly implement chmod in buildah this addition is needed.

Part of WIP:
https://github.com/containers/buildah/pull/3004

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -153,6 +153,10 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 			chown = strings.TrimPrefix(arg, "--chown=")
 		case strings.HasPrefix(arg, "--chmod="):
 			chmod = strings.TrimPrefix(arg, "--chmod=")
+			_, err = strconv.ParseUint(chmod, 8, 32)
+			if err != nil {
+				return errChmodConversion(chmod)
+			}
 		default:
 			return fmt.Errorf("ADD only supports the --chmod=<permissions> and the --chown=<uid:gid> flag")
 		}
@@ -185,6 +189,10 @@ func dispatchCopy(b *Builder, args []string, attributes map[string]bool, flagArg
 			chown = strings.TrimPrefix(arg, "--chown=")
 		case strings.HasPrefix(arg, "--chmod="):
 			chmod = strings.TrimPrefix(arg, "--chmod=")
+			_, err = strconv.ParseUint(chmod, 8, 32)
+			if err != nil {
+				return errChmodConversion(chmod)
+			}
 		case strings.HasPrefix(arg, "--from="):
 			from = strings.TrimPrefix(arg, "--from=")
 		default:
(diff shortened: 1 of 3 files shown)


jmguzik:
@rhatdan FYI

jmguzik:
@rhatdan @nalind @smarterclayton could you mark this change as ok to test?

jmguzik:
@nalind since all jobs succeded, and you responded do fast, could I ask you to review this one?

dispatchers_test.go
```diff
@@ -2,11 +2,12 @@ package imagebuilder
 
 import (
 	"errors"
-	"github.com/containerd/containerd/platforms"
 	"reflect"
 	"sort"
 	"testing"
 
+	"github.com/containerd/containerd/platforms"
+
```
rhatdan, line 10: Nit: remove blank line.

rhatdan (approved):
Small nit, but LGTM

dispatchers_test.go
jmguzik: Line removed, thanks.

nalind:
Tests that use `--chmod=888` look kind of iffy, in that they seem to only be testing that 888 doesn't get interpreted as 777 or 755, otherwise it looks correct.

jmguzik:
@nalind What would be your suggestion? Change them (could you propose something?) or get rid of them?

rhatdan:
@jmguzik Doesn't the `888` value or `bogus` trigger a real error?

jmguzik:
> @jmguzik Doesn't the `888` value or `bogus` trigger a real error?

No, right now it acts exactly like chown implementation. Does not check anything, just forwards.
In fact 888 could be replaced by any other string and if checked in params, and it would be validated as correct one. I just followed chown implementation of bad values and chose 888 to reflect it in test. Could be any other string.
In my understanding it is not the place to check, however if you would like to see this check here please let me know. It would land probably in add and copy functions in builder.

I think what @nalind meant is just to remove 888 to replace it with any other value from range to remve "iffy look". From the testing perspective, nothing will change. The same applies to chown impl.

rhatdan:
Ok do that and I will merge.

jmguzik:
> Ok do that and I will merge.

Done

dispatchers.go
```diff
@@ -179,13 +187,19 @@ func dispatchCopy(b *Builder, args []string, attributes map[string]bool, flagArg
 		switch {
 		case strings.HasPrefix(arg, "--chown="):
 			chown = strings.TrimPrefix(arg, "--chown=")
+		case strings.HasPrefix(arg, "--chmod="):
+			chmod = strings.TrimPrefix(arg, "--chmod=")
+			_, err = strconv.ParseUint(chmod, 8, 32)
+			if err != nil {
+				return errChmodConversion(chmod)
+			}
```
TomSweeneyRedHat, line 195: nit, you might consider renaming errChmodConversion to something like checkChmodConversion then putting the ParseUnit() call into and then return err or nil

TomSweeneyRedHat (commented):
One nit/suggestion.
Otherwise LGTM, but would like a head nod from @nalind

dispatchers.go
jmguzik: This is exactly what I had (even the same name), but then spotted err bla bla... funtions at the bottom and decided to follow this convention. Anyway, I will fix it right away.

jmguzik:
Done. Added one more check in tests to prevent a problem with null.

PR #185, opened 2021-02-22
README.txt: correct the suggested invocation for conformance tests

nalind:
The conformance tests shouldn't be run against a single source file, but rather the package that contains them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,8 @@ Example of usage from OpenShift's experimental `dockerbuild` [command with mount
 ## Run conformance tests (very slow):
 
 ```
+docker rmi busybox; docker pull busybox
+docker rmi centos:7; docker pull centos:7
 chmod -R go-w ./dockerclient/testdata
-go test ./dockerclient/conformance_test.go -tags conformance -timeout 30m
+go test ./dockerclient -tags conformance -timeout 30m
 ```


README.md
```diff
@@ -102,6 +102,8 @@ Example of usage from OpenShift's experimental `dockerbuild` [command with mount
 ## Run conformance tests (very slow):
 
 ```
+docker rmi busybox; docker pull busybox
+docker rmi centos:7; docker pull centos:7
 chmod -R go-w ./dockerclient/testdata
-go test ./dockerclient/conformance_test.go -tags conformance -timeout 30m
+go test ./dockerclient -tags conformance -timeout 30m
```
TomSweeneyRedHat, line 108: I think this will run both the conformance and copyinfo tests.  Is that intended?

nalind: Kind of unavoidable, I think, unless we add a build tag to exclude `copy_test.go` from conformance test runs:  when we tell `go test` to test a particular file, it doesn't seem to be resolving references to symbols from other source files in the same directory as the one that holds the specified file.

TomSweeneyRedHat:
LGTM

rhatdan:
Since I was the victim of this  /lgtm

PR #186, opened 2021-02-22
Bump to v1.2.0

rhatdan:


rhatdan:
@nalind @TomSweeneyRedHat PTAL

nalind:
If we're calling this 1.2.0, does the development version start claiming it's 1.3-dev?  Otherwise LGTM.

rhatdan:
We usually only bump the Z access for -dev. rather then the X.  X should only get bumped when there are new features line the chmod stuff.

TomSweeneyRedHat:
Should we be creating a release branch for imagebuilder like we have for the other projects?  The only one we have is an OCP 3.11 one.

nalind:
> We usually only bump the Z access for -dev. rather then the X. X should only get bumped when there are new features line the chmod stuff.

For v1.2.0, X=1, Y=2, Z=0.  The previous release tag was v1.1.8, so this was a bump on Y.

rhatdan:
I miswrote.

We usually only bump the Z access for -dev. rather then the Y. Y should only get bumped when there are new features line the chmod stuff.  

The X Should only be bumped for breaking changes (After 2).

Since we added new functionality we bump the Y.

PR #188, opened 2021-03-22
dockerclient: implement --chown for ADD/COPY

nalind:
Fixes #187.

PR #189, opened 2021-03-24
dockerclient/testdata/copychown/Dockerfile: remove an irrelevant instruction

nalind:
Running "ls" on the /usr/bin/script provided by util-linux doesn't contribute anything to this test, so don't bother with that.


nalind:
h/t to @TomSweeneyRedHat for asking about it in #188.

PR #190, opened 2021-04-01
COPY: treat sources named "" as "*"

### LINKED ISSUE #181: COPY does not handle empty arguments well
```sh
cat <<EOF >Dockerfile
FROM centos
COPY Dockerfile "" /tmp/
RUN ls /
RUN ls /tmp
EOF
imagebuilder .
```
This will create a file `/tmpDockerfile`. in `/` `docker build` ignores the empty string.

Use case: An empty `ARG` variable that could be used to copy extra files in.

nalind:
If a source for COPY is "", treat it as though it were "*".  Fixes #181.


TomSweeneyRedHat:
LGTM
and happy green test buttons.

PR #191, opened 2021-04-01
COPY: treat absolute source paths as being relative to the contextdir

### LINKED ISSUE #139: "COPY /blah/blub /path" does not work same way as docker
The COPY command does not work for the following case - source files are not interpreted as relativ to the current workspace as stated in the original docker refence:

https://docs.docker.com/engine/reference/#copy

> Multiple \<src\> resources may be specified but the paths of files and directories will be _interpreted as relative_ to the source of the context of the build.

Example workspace directory:
```
Dockerfile
firstdir/seconddir/dir-a/file-a
firstdir/seconddir/dir-b/file-b
```

Dockerfile content:
```
FROM alpine:3.10
COPY /firstdir/seconddir /var
RUN ls -la /var && ls -la /var/firstdir
RUN ls -la /var/dir-a
```

The original docker builder copies the content of the firstdir/seconddir to the image, because all paths are relative to the context. `imagebuilder` failes with an error message saying `stat seconddir: no such file or directory`

Changing the COPY line to the following (removing first slash)
```
COPY firstdir/seconddir /var
```
Imagebuilder works as expected.

nalind:
Source paths for COPY that look like absolute paths are relative to the context directory, so treat them as such.  Fixes #139.


rhatdan:
LGTM
@TomSweeneyRedHat @vrothberg @saschagrunert PTAL

nalind:
Added an additional test to cover the second case from #139.

TomSweeneyRedHat:
LGTM

vrothberg (commented):
LGTM

saschagrunert (approved):
LGTM

Issue #187, opened 2021-03-09
COPY --chown does not work right


sosiouxme:
https://bugzilla.redhat.com/show_bug.cgi?id=1937087

```
$ cat > Dockerfile << EOF
FROM registry.access.redhat.com/ubi8:latest
COPY --chown=1001:0 myfile /myfile
RUN ls -l /myfile
EOF

$ touch myfile

$ imagebuilder --tag yolo:latest .
--> FROM registry.access.redhat.com/ubi8:latest as 0
--> COPY --chown=1001:0 myfile /myfile
--> RUN ls -l /myfile
-rw-rw-r--. 1 root root 0 Mar  9 19:01 /myfile
--> Committing changes to yolo:latest ...
--> Done

# ^ Notice how the file is owned by root when using imagebuilder
# v Notice how the file has the expected ownership when using podman

$ podman build --no-cache -t yolo:latest .
STEP 1: FROM registry.access.redhat.com/ubi8:latest
STEP 2: COPY --chown=1001:0 myfile /myfile
--> ea573ccd64f
STEP 3: RUN ls -l /myfile
-rw-rw-r--. 1 1001 root 0 Mar  9 19:01 /myfile
STEP 4: COMMIT yolo:latest
--> 4a1c709a0c0
4a1c709a0c04b310cdbf45fd8791a534e2c54c320634da2c5c153b74b87fa67b
```

PR #193, opened 2021-04-13
Add mounts to dispatch run

ashley-cui:
parse build mounts RUN --mount=...
the RUN command can take flags from the dockerfile.

needed for https://github.com/containers/buildah/issues/1684

Signed-off-by: Ashley Cui <acui@redhat.com>

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -37,6 +37,8 @@ type Copy struct {
 type Run struct {
 	Shell bool
 	Args  []string
+	// Mounts are mounts specified through the --mount flag inside the Containerfile
+	Mounts []string
 }
 
 type Executor interface {
@@ -67,7 +69,7 @@ func (logExecutor) Copy(excludes []string, copies ...Copy) error {
 }
 
 func (logExecutor) Run(run Run, config docker.Config) error {
-	log.Printf("RUN %v %t (%v)", run.Args, run.Shell, config.Env)
+	log.Printf("RUN %v %v %t (%v)", run.Args, run.Mounts, run.Shell, config.Env)
 	return nil
 }
 
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -306,7 +306,26 @@ func run(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 
 	args = handleJSONArgs(args, attributes)
 
-	run := Run{Args: args}
+	var mounts []string
+	userArgs := mergeEnv(envMapAsSlice(b.Args), b.Env)
+	for _, a := range flagArgs {
+		arg, err := ProcessWord(a, userArgs)
+		if err != nil {
+			return err
+		}
+		switch {
+		case strings.HasPrefix(arg, "--mount="):
+			mount := strings.TrimPrefix(arg, "--mount=")
+			mounts = append(mounts, mount)
+		default:
+			return fmt.Errorf("RUN only supports the --mount flag")
+		}
+	}
+
+	run := Run{
+		Args:   args,
+		Mounts: mounts,
+	}
 
 	if !attributes["json"] {
 		run.Shell = true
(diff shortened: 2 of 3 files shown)


nalind:
/ok-to-test
LGTM, thanks!

dockerclient/client.go
```diff
@@ -694,6 +694,10 @@ func (e *ClientExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error
 // the user command into a shell and perform those operations before. Since RUN
 // requires /bin/sh, we can use both 'cd' and 'export'.
 func (e *ClientExecutor) Run(run imagebuilder.Run, config docker.Config) error {
+	if len(run.Mounts) > 0 {
+		return fmt.Errorf("RUN --mount not supported")
```
rhatdan, line 698: So we are adding support just to throw an error if the user uses it?

ashley-cui: Took another look, add support! It should use the same code as what I'm writing in buildah so it'll be fairly easy

ashley-cui: Wait sorry, took another look at line 767 
if err := e.Client.StartExec(exec.ID, docker.StartExecOptions{...}
From my understanding, imagebuilder run uses docker.StartExecOptions, and that doesn't support the --mounts flag. I could add the mounts to `e.TransientMounts` but that means the mount will be visible for other steps in the build, which is what we don't want when we use the --mount flag with a RUN. 

I'm adding support for this because we can do what I said above in buildah, since we can add a mounts field to buildah's builder.RunOptions.

ashley-cui: Let me know though if there's a way around it or if I'm reading it wrong

ashley-cui: update: I might have found a way, let me take another look, sorry for the noise

nalind: > So we are adding support just to throw an error if the user uses it?

I find it preferable to adding the support for parsing the flag, which we need to do in order to do something with the flag in buildah, and then just ignoring it in the dockerclient.  That'd be how we got #187.

ashley-cui: Sure, I can just remove the error

builder.go
```diff
@@ -37,6 +37,8 @@ type Copy struct {
 type Run struct {
 	Shell bool
 	Args  []string
+	// Mounts are mounts specified through the --mount flag inside the dockerfile
```
TomSweeneyRedHat, line 40: Only if you've other changes
```suggestion
	// Mounts are mounts specified through the --mount flag inside the Dockerfile
```

builder.go
```diff
@@ -67,7 +69,7 @@ func (logExecutor) Copy(excludes []string, copies ...Copy) error {
 }
 
 func (logExecutor) Run(run Run, config docker.Config) error {
-	log.Printf("RUN %v %t (%v)", run.Args, run.Shell, config.Env)
+	log.Printf("RUN %v %v %t (%v)", run.Args, run.Mounts, run.Shell, config.Env)
```
TomSweeneyRedHat, line 72: This is probably more a question for @nalind and @rhatdan and I don't think it's a problem, but I don't know that it's not.  Any chance somebody might be parsing this log output already and having the mounts added in the middle rather than the end would mess that up?

builder.go
rhatdan: Better yet call it Containerfile

builder.go
rhatdan: I would doubt people are scanning these log messages.

TomSweeneyRedHat:
LGTM

builder.go
nalind: imagebuilder doesn't look for files with that name.

builder.go
nalind: I don't expect it to create problems, either.  We already tweaked the log format for COPY in #184, and I haven't heard anything yet.

PR #195, opened 2021-07-29
Handle COPY --from=${number}

nalind:
Handle COPY --from=${stage} where the referred-to stage is referred to by its order in the build, and add a couple of test Dockerfiles that test COPY --from=${stage}, where ${stage} is a number.  Conformance tests don't currently handle multi-stage builds, but we add a couple of samples for manual smoke testers.

### DIFF
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -138,14 +138,19 @@ func (e *ClientExecutor) DefaultExcludes() error {
 }
 
 // WithName creates a new child executor that will be used whenever a COPY statement
-// uses --from=NAME.
-func (e *ClientExecutor) WithName(name string) *ClientExecutor {
+// uses --from=NAME or --from=POSITION.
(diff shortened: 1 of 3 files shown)


dockerclient/client.go
```diff
@@ -163,6 +168,7 @@ func (e *ClientExecutor) WithName(name string) *ClientExecutor {
 
 	child := &copied
 	e.Named[name] = child
+	e.Named[fmt.Sprintf("%d", position)] = child
```
rhatdan, line 171: Why not use strconf.Itoa() it is supposed to be more efficient.

rhatdan (approved):
LGTM other then one question.

dockerclient/client.go
nalind: Only that I forgot about that function.  Changing it.

PR #196, opened 2021-07-30
Migrate to Go modules

nalind:
Newer versions of Go refuse to build imagebuilder without a go.mod.  So let's migrate from the older vendor.conf to modules with the help of `go mod init`.


nalind:
Okay, we're down to the differences in which files `vndr` and `go mod vendor` keep, except for `google.golang.org/grpc`.

TomSweeneyRedHat:
@vrothberg PTAL

TomSweeneyRedHat:
LGTM
Not required for this PR, but it might be nice to have a little blurb somewhere on how to vendor.  I know I reference the blurb at the bottom of the install.md on buildah all the time.

PR #197, opened 2021-08-02
Use -mod=vendor when available, update dependencies

nalind:
Use `-mod=vendor` when the compiler supports it, and update github.com/containers/storage to get around security problems we would have had in the very old version that we were previously using.  This pulls in newer versions of just about everything else.


nalind:
I expect us to block on https://github.com/openshift/release/pull/20742.

.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: go
 
 go:
-  - "1.9"
-  - "1.10"
+  - "1.15"
+  - "1.16"
```
TomSweeneyRedHat, line 5: Can we go to: golang 1.16.6, golang 1.15.14, there's a new CVE in versions less than that.  1.17 is also free of issues CVE wise, but I'm not sure if it's available everywhere yet.

Makefile
```diff
@@ -1,11 +1,26 @@
+export GOPROXY=https://proxy.golang.org
+
+GO := go
+GO110 := 1.10
```
TomSweeneyRedHat, line 4: should this be bumped too?

TomSweeneyRedHat:
Tests are not so hip

.travis.yml
nalind: "1.15.x" and "1.16.x" should get us what we want.  1.17 is not yet released.

Makefile
nalind: Hmm, don't think we actually need that.  Removing it.

nalind:
> Tests are not so hip

I expected us to block on https://github.com/openshift/release/pull/20742.

PR #198, opened 2021-08-02
archiveFromDisk: inline archives that are being ADDed

### LINKED ISSUE #194: ADD should unpack archives
My Dockerfile has a "ADD" directive, which should unpack the tar.gz files in the image and it works fine for a "regular" build. But if I change the container.yaml to use `image_build_method: imagebuilder` instead, the "ADD" directive copies the tar.gz without unpacking it.

Reproducer:

container.yaml:
```yaml
compose:
  pulp_repos: true
image_build_method: imagebuilder
platforms:
  only:
    - x86_64
    - s390x
    - ppc64le
```

Dockerfile
```Dockerfile
FROM registry.redhat.io/ubi8/ubi
ADD *.tar.gz /packages/
COPY setup.sh /
RUN /setup.sh
```

setup.sh
```bash
echo "LIST PACKAGES:"
ls -l packages
```

It will list tar.gz files. Removing the `image_build_method` node from `container.yaml` will cause directories to be listed.

Reference: https://docs.docker.com/engine/reference/builder/#add

> If < src > is a local tar archive in a recognized compression format (identity, gzip, bzip2 or xz) then it is unpacked as a directory. Resources from remote URLs are not decompressed. When a directory is copied or unpacked, it has the same behavior as tar -x, the result is the union of:
>
>    1. Whatever existed at the destination path and
>    2. The contents of the source tree, with conflicts resolved in favor of “2.” on a file-by-file basis.

nalind:
Remove the attempt to handle archives from ClientExecutor.Archive(), which couldn't handle sources specified using globbing patterns, and move that logic into archiveFromDisk(), which handles the globbing.

Fixes #194.

### DIFF
--- a/dockerclient/archive.go
+++ b/dockerclient/archive.go
@@ -16,6 +16,7 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/fileutils"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/ioutils"
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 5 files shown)


nalind:
Added another conformance test.

dockerclient/archive.go
```diff
@@ -182,9 +183,93 @@ func archiveFromDisk(directory string, src, dst string, allowDownload bool, excl
...
+					if renamed, ok := options.RebaseNames[includeFile]; ok {
+						hdr.Name = strings.TrimSuffix(renamed, includeFile) + hdr.Name
+						if hdr.Typeflag == tar.TypeLink {
+							hdr.Linkname = strings.TrimSuffix(renamed, includeFile) + hdr.Linkname
+						}
+					}
+					tw.WriteHeader(hdr)
+					_, err = io.Copy(tw, tr)
+					if err != nil {
+						break
```
rhatdan, line 223: Do we care about this err?  Seems like it could just get lost?  Should it at least be logged somewhere?

dockerclient/archive.go
```diff
@@ -182,9 +183,93 @@ func archiveFromDisk(directory string, src, dst string, allowDownload bool, excl
...
+					}
+					hdr, err = tr.Next()
+				}
+				dc.Close()
```
rhatdan, line 227: Why not defer these dc.Close() and f.Close()

dockerclient/archive.go
```diff
@@ -182,9 +183,93 @@ func archiveFromDisk(directory string, src, dst string, allowDownload bool, excl
...
+			hdr, err := tr.Next()
+			for err == nil {
+				tw.WriteHeader(hdr)
+				_, err = io.Copy(tw, tr)
+				if err != nil {
+					break
+				}
+				hdr, err = tr.Next()
+			}
+			rc.Close()
```
rhatdan, line 256: Why not defer?

dockerclient/client.go
```diff
@@ -1079,27 +1079,29 @@ func (e *ClientExecutor) Archive(fromFS bool, src, dst string, allowDownload boo
...
-			h.Name = filepath.Join(dst, h.Name)
-			if h.Typeflag == tar.TypeLink {
-				h.Linkname = filepath.Join(dst, h.Linkname)
+	if false {
+		// if the source is an archive, extract it under the destination directory
+		srcFullPath := filepath.Join(e.Directory, src)
+		if allowDownload && isArchivePath(srcFullPath) {
+			f, err := os.Open(srcFullPath)
+			if err != nil {
+				return nil, nil, fmt.Errorf("error opening %q: %v", srcFullPath, err)
```
rhatdan, line 1088: This will cause a stutter.

dockerclient/client.go
```diff
@@ -1079,27 +1079,29 @@ func (e *ClientExecutor) Archive(fromFS bool, src, dst string, allowDownload boo
...
+				h.Name = filepath.Join(dst, h.Name)
+				if h.Typeflag == tar.TypeLink {
+					h.Linkname = filepath.Join(dst, h.Linkname)
+				}
+				return nil, false, false, nil
+			}
+			filtered, err := transformArchive(f, true, transform)
+			if err != nil {
+				f.Close()
+				return nil, nil, fmt.Errorf("error transforming archive %q: %v", srcFullPath, err)
```
rhatdan, line 1101: Potential stutter?

rhatdan (approved):

dockerclient/archive.go
nalind: I didn't want to build up a backlog of descriptors and pipes to close, but I'll change it to add a scope there.

dockerclient/archive.go
nalind: No good reason.  Changing it.

dockerclient/client.go
nalind: Tweaking the format string to "open %q: %v".

dockerclient/client.go
nalind: Not sure, but replacing "error" with "while".

dockerclient/client.go
rhatdan: The open failure will include an error with srcFullPath mentioned, if you wrap with srcFullPath, then you will print it twice.

nalind: That whole section's been removed, so I'm going to call it fixed.

dockerclient/client.go
nalind: Removed this, too.

TomSweeneyRedHat:
LGTM

PR #204, opened 2021-10-19
calcCopyInfo: also trim leading path on rootPath

nalind:
calcCopyInfo trims a leading "./" or "/./" from one of its parameters, but not the other, which can break COPY instructions when the build context location is specified using a path that starts with "./".


TomSweeneyRedHat:
LGTM

PR #205, opened 2022-01-10
Avoid panic when writing UIDs/GIDs > 0x1fffff

### LINKED ISSUE #200: Imagebuilder gets stuck while copying content between stages with --chown for a user with specific UID
One of our OSBS users accidentally found a bug in the imagebuilder. Their multistage builds started timing out on Dockerfile instruction:
```
COPY --from=builder --chown=submariner:submariner /tmp/files /opt/files/
```

RCA revealed that the real issue is s specific UID with a **too long value** for a new user. If this value is changed from 1001010000 to e.g. 1005 or if a user is created without a specific UID, everything works fine. 

OSBS system uses following version of imagebuilder: 
```
lkolacek@archlinux ~ $ sudo podman run -it registry-proxy.engineering.redhat.com/rh-osbs/rcm-buildroot:4.0-28 rpm -q imagebuilder
imagebuilder-1.2.1-1.el8eng.x86_64
```

Dockerfile with reproducer:
```
lkolacek@archlinux osbs-test-sandwich (cachito-copy-chown $%=) $ cat Dockerfile 
FROM registry.redhat.io/ubi8/ubi-minimal@sha256:31ccb79b1b2c2d6eff1bee0db23d5b8ab598eafd6238417d9813f1346f717c11 AS builder

LABEL com.redhat.component="osbs-test-sandwich-container" \
      name="osbs-test/multiarch-base-test" \
      version="1.0.multiarch.base.test"

# Copy the upstream sources from cachito integration
COPY $REMOTE_SOURCE $REMOTE_SOURCE_DIR

COPY files /tmp/files

CMD ["exit", "0"]

FROM registry.redhat.io/ubi8/ubi-minimal@sha256:31ccb79b1b2c2d6eff1bee0db23d5b8ab598eafd6238417d9813f1346f717c11

LABEL com.redhat.component="osbs-test-sandwich-container" \
      name="osbs-test/cachito-copy-chown" \
      version="1.0.cachito.copy.chown"

RUN microdnf update && \
    microdnf -y install --nodocs shadow-utils && \
    microdnf clean all

RUN adduser -r -l -u 1001010000 -G root submariner
COPY --from=builder --chown=submariner:submariner /tmp/files /opt/files/

CMD ["/usr/bin/hello_world"]
```

nalind:
The USTAR archive format can't encode UIDs or GIDs that can't be represented in 21 (7 * 3) bits, which creates errors when we modify headers to include such values.  When we do that, try to force a conversion to the PAX format, which can handle that.

This requires Go 1.10 or later, as the constants which need to be set in the header to indicate the format which should be used weren't exported before that.

Fixes #200.

### DIFF
--- a/dockerclient/archive.go
+++ b/dockerclient/archive.go
@@ -576,8 +576,8 @@ func (m *archiveMapper) Filter(h *tar.Header, r io.Reader) ([]byte, bool, bool, 
 					rehdr, err = tr2.Next()
 				}
 				if err != nil {
-					return nil, false, true, fmt.Errorf("needed to create %q as a hard link to %q, but got error refetching %q: %v", h.Name, h.Linkname, h.Linkname, err)
 					pr.Close()
+					return nil, false, true, fmt.Errorf("needed to create %q as a hard link to %q, but got error refetching %q: %v", h.Name, h.Linkname, h.Linkname, err)
 				}
 				buf, err := ioutil.ReadAll(pr)
 				pr.Close()
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -116,7 +116,7 @@ type ClientExecutor struct {
 	Volumes *ContainerVolumeTracker
 }
 
-// NotAuthFn can be used for AuthFn when no authentication is required in Docker.
+// NoAuthFn can be used for AuthFn when no authentication is required in Docker.
 func NoAuthFn(string) ([]dockertypes.AuthConfig, bool) {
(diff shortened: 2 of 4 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
Changes LGTM
would be nice to have an added test if possible

flouthoc:
I am Late. But since right interface is implemented for old golang versions. LGTM

PR #206, opened 2022-01-10
Authenticate to registries when pulling images

### LINKED ISSUE #144: imagebuilder fails with `unauthorized: authentication required` when it should not (proper auths in ~/.docker/config.json)
These works (several repositories, [here's an example](https://github.com/openshift/cluster-config-operator) ):
`buildah bud -f Dockerfile.rhel7 -t myimage:local .`
`docker build -f Dockerfile.rhel7 -t myimage:local .`

This fails:
`imagebuilder --allow-pull -t myimage:local -f ./Dockerfile.rhel7 .`
Fails with:
```console
$ imagebuilder --allow-pull -t quay.io/sallyom/myimage:local -f ./Dockerfile.rhel7 .
--> Image registry.svc.ci.openshift.org/ocp/builder:golang-1.12 was not found, pulling ...
unable to pull image (from: registry.svc.ci.openshift.org/ocp/builder, tag: golang-1.12): unauthorized: authentication required
```
Workaround is to pre-pull with docker whatever you need:
`docker pull registry.svc.ci.openshift.org/ocp/4.2:base`
`docker pull registry.svc.ci.openshift.org/ocp/builder:golang-1.12`
(has to be in docker storage rather than buildah,podman)

nalind:
When we attempt to pull images, use authentication secrets, if we can find some that match.  We special-case docker.io and locations that end in .docker.io to also check for authentication information tagged for either docker.io, index.docker.io, or https://index.docker.io/v1/.

Fixes #144.

### DIFF
--- a/cmd/imagebuilder/imagebuilder.go
+++ b/cmd/imagebuilder/imagebuilder.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/docker/distribution/reference"
 	dockertypes "github.com/docker/docker/api/types"
 	docker "github.com/fsouza/go-dockerclient"
 	"k8s.io/klog"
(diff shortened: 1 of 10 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
Changs LGTM
any way to add a test or two?

rhatdan:
Add tests in a subsequent PR.

cmd/imagebuilder/imagebuilder.go
```diff
@@ -87,7 +88,58 @@ func main() {
...
+					return []dockertypes.AuthConfig{{
+						Username:      authConfig.Username,
+						Password:      authConfig.Password,
+						Email:         authConfig.Email,
+						ServerAddress: authConfig.ServerAddress,
+					}}, true
+				}
+				if domain == "docker.io" || strings.HasSuffix(domain, ".docker.io") {
+					var auths []dockertypes.AuthConfig
+					for _, aka := range []string{"docker.io", "index.docker.io", "https://index.docker.io/v1/"} {
```
flouthoc, line 123: @nalind I think single auth credential would work for all if them `"docker.io", "index.docker.io", "https://index.docker.io/v1/"`. If individual entries for each of them is found it would return with three similar auth credentials.

Instead of append we could a `return` I think.

flouthoc:
I am late. PR looks good to me just one doubt and its a non-blocker @nalind PTAL

cmd/imagebuilder/imagebuilder.go
nalind: Yes, the caller (ClientExecutor.LoadImage) will try them all, stopping after any of them succeed.

PR #207, opened 2022-01-11
COPY --chown: add missing parent directories

nalind:
When we let the daemon implicitly create not-yet-created parent directories when we COPY in content that is to be owned by a non-root user, the daemon doesn't know that newly-created directories also need to be owned by that non-root user, so it creates them owned by root, which is wrong.

When we're using COPY with --chown, check for the destination path for any copy being missing, and add any missing path components as directories.  If the path is a destination file, it'll be overwritten by the file.  This is pretty expensive, though, so don't do it unless we have to.

In `TestConformanceInternal/copy chown`, test `--chown` with a value larger than we can represent in 21 bits, so that `forceHeaderFormat()` will be invoked, to test #205.

### DIFF
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -13,6 +13,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -131,6 +132,9 @@ func NewClientExecutor(client *docker.Client) *ClientExecutor {
 	}
 }
 
+// DefaultExcludes reads the default list of excluded file patterns from the
+// context directory's .containerignore file if it exists, or from the context
+// directory's .dockerignore file, if it exists.
 func (e *ClientExecutor) DefaultExcludes() error {
 	var err error
 	e.Excludes, err = imagebuilder.ParseDockerignore(e.Directory)
@@ -638,6 +642,10 @@ func (e *ClientExecutor) Preserve(path string) error {
 }
 
 func (e *ClientExecutor) EnsureContainerPath(path string) error {
+	return e.createOrReplaceContainerPathWithOwner(path, 0, 0)
+}
+
+func (e *ClientExecutor) createOrReplaceContainerPathWithOwner(path string, uid, gid int) error {
 	createPath := func(dest string) error {
 		var writerErr error
 		if !strings.HasSuffix(dest, "/") {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

nalind:
Added a case where the destination is a top-level directory, and fixed it.

nalind:
Revised to use a helper function I hadn't noticed before.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
FWIW, I'm a bit worried about the expense.  @flouthoc any thoughts?

dockerclient/client.go
```diff
@@ -872,6 +882,20 @@ func (e *ClientExecutor) CopyContainer(container *docker.Container, excludes []s
 		}
 		return *value, nil
 	}
+	findMissingParents := func(dest string) (parents []string, err error) {
+		destParent := filepath.Clean("/" + dest)
```
flouthoc, line 886: Isn't this trying to treat `relative path` as an `absolute` one.

After `WORKDIR`  change user could pass a relative path. But `destParent` is cleaning the path and assuming it as an absolute path.

```dockerfile
FROM alpine
RUN mkdir -p test/hello/world

WORKDIR test/hello/world

COPY hello ../hello/hello

WORKDIR ../hello
RUN ls -a

``` 

So `filepath.Clean("/"+"./somedir")` would make it `/somedir`

It would be okay if we are sure that `c.Dest` in `findMissingParents(c.Dest)` is a fully resolved path not `relative`.

dockerclient/testdata/copychown/Dockerfile
```diff
@@ -1,10 +1,14 @@
...
-COPY --chown=bin    script /usr/bin/script.bin
-COPY --chown=lp     script /usr/bin/script.lp
-COPY --chown=3      script script2 /usr/local/bin/
+COPY --chown=1:2     script /usr/bin/script.12
+COPY --chown=1:adm   script /usr/bin/script.1-adm
+COPY --chown=1       script /usr/bin/script.1
+COPY --chown=lp:adm  script /usr/bin/script.lp-adm
+COPY --chown=2:mail  script /usr/bin/script.2-mail
+COPY --chown=2       script /usr/bin/script.2
+COPY --chown=bin     script /usr/bin/script.bin
```
flouthoc, line 8: It would be nice if test has some `absolute` paths as well.

dockerclient/client.go
nalind: The callers ensure that it is, so I'll remove the bit that prepends a "/" before cleaning.

dockerclient/testdata/copychown/Dockerfile
nalind: Adding.

flouthoc:
@nalind Thanks ! LGTM 
/lgtm
/approve

PR #211, opened 2022-01-26
dockerclient.ClientExecutor.Stages(): commit base stages

### LINKED ISSUE #209: ENV does not inherit through multiple layers
Given the following test case:
``` Dockerfile
FROM registry.redhat.io/ubi8:latest AS base
ENV FOO=foo
RUN echo FOO=$FOO

FROM base AS devel
ENV BAR=bar
RUN echo FOO=$FOO BAR=$BAR

FROM devel AS devel2
RUN echo FOO=$FOO BAR=$BAR

FROM devel2 AS devel3
RUN echo FOO=$FOO BAR=$BAR
```
With both docker and podman, the variables persist through multiple layers of inheritance:
```
$ podman build -f Dockerfile.test .
[1/4] STEP 1/3: FROM registry.redhat.io/ubi8:latest AS base
[1/4] STEP 2/3: ENV FOO=foo
--> Using cache 517c793b4782c1d4119961ab8b3644cb2318f44bb666ffa085de1cf0be25f059
--> 517c793b478
[1/4] STEP 3/3: RUN echo FOO=$FOO
--> Using cache 13e4677bfaaf33b4fb2ca4232d99f2d65975145db458126cfef2274679ab0719
--> 13e4677bfaa
[2/4] STEP 1/3: FROM 13e4677bfaaf33b4fb2ca4232d99f2d65975145db458126cfef2274679ab0719 AS devel
[2/4] STEP 2/3: ENV BAR=bar
--> Using cache ae024ad2983b8e2fe9cb49cafa798cdca656635a646f04ea716aaa597c794a81
--> ae024ad2983
[2/4] STEP 3/3: RUN echo FOO=$FOO BAR=$BAR
--> Using cache 92a4f207c8fbe090b828b397a0b50738950f71e8d3e9884a4a27a961d917afa5
--> 92a4f207c8f
[3/4] STEP 1/2: FROM 92a4f207c8fbe090b828b397a0b50738950f71e8d3e9884a4a27a961d917afa5 AS devel2
[3/4] STEP 2/2: RUN echo FOO=$FOO BAR=$BAR
FOO=foo BAR=bar
--> f542605d956
[4/4] STEP 1/2: FROM f542605d956a5ccc92578c3a03a45c4b58d36405f195eff6d31667ee75a197f8 AS devel3
[4/4] STEP 2/2: RUN echo FOO=$FOO BAR=$BAR
FOO=foo BAR=bar
[4/4] COMMIT
--> ad65536fb76
ad65536fb76248a3a45fd21478ec23b37d532b225846bc4d009d93b07fa3f149
```
However, with imagebuilder, the variables do not persist through more than one layer of inheritance:
```
--> FROM registry.redhat.io/ubi8:latest as base
--> ENV FOO=foo
--> RUN echo FOO=$FOO
FOO=foo
--> FROM base as devel
--> ENV BAR=bar
--> RUN echo FOO=$FOO BAR=$BAR
FOO=foo BAR=bar
--> FROM devel as devel2
--> RUN echo FOO=$FOO BAR=$BAR
FOO= BAR=bar
--> FROM devel2 as devel3
--> RUN echo FOO=$FOO BAR=$BAR
FOO= BAR=
--> Committing changes ...
--> Done
```
The real-world case where I saw this is [quay](https://github.com/quay/quay/blob/master/Dockerfile), where variables set in base do not make it through to `build-python` (where `build` is the intermediate layer between them).

nalind:
When we attempt to use the result of a previous stage as a base image for a later stage, instead of just continuing to use the previous stage's container, commit the previous stage's container to create an image which we'll then use to create a new container for use in the stage.

Builds which looked like this:
```Dockerfile
  FROM $ExternalBaseImage AS commonBase
  (install general updates)
  FROM commonBase AS builder
  (install build tooling and compile code)
  FROM commonBase AS final
  (copy compiled code from "builder" stage)
```
would otherwise end up producing an image from its final stage which contained all of the changes made in the middle stage, which is not compatible with what `docker build` does.

Note that `dockerclient.ClientExecutor.Build()` now saves the committed image's ID in its new `Committed.ID` field rather than the `Image.ID` field, which previously either contained either the ID of a base image or of the committed image.

Fixes #209.

### DIFF
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -100,6 +100,9 @@ type ClientExecutor struct {
 	// as a base for this build. Otherwise the FROM value from the
 	// Dockerfile is read (will be pulled if not locally present).
 	Image *docker.Image
+	// Committed is optional and is used to track a temporary image, if one
+	// was created, that was based on the container as its stage ended.
+	Committed *docker.Image
 
 	// AuthFn will handle authenticating any docker pulls if Image
 	// is set to nil.
@@ -170,6 +173,7 @@ func (e *ClientExecutor) WithName(name string, position int) *ClientExecutor {
 	copied.Deferred = nil
 	copied.Image = nil
 	copied.Volumes = nil
+	copied.Committed = nil
 
 	child := &copied
 	e.Named[name] = child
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@flouthoc PTAL

flouthoc (approved):
LGTM

dockerclient/client.go
```diff
@@ -437,7 +464,7 @@ func (e *ClientExecutor) Commit(b *imagebuilder.Builder) error {
 		return fmt.Errorf("unable to commit build container: %v", err)
 	}
 
-	e.Image = image
```
flouthoc, line 440: @nalind Is this a breaking change for library users ? Custom code which is relying on `ClientExecutor.Image` would now get a `nil` or incorrect value cause we are moving committed image to `ClientExecutor.Committed` 

Just a small concern although I doubt if anybody is going to be using `ClientExecutor` directly.

flouthoc:
Just one doubt in above comment otherwise LGTM.

dockerclient/client.go
nalind: It would, but since there was no indication of whether or not `ClientExecutor.Image` referred to the base image, or the committed image (the caller had to keep track of whether it had committed an image or not), it wasn't ideal.  The only library consumer I know of is the release-3.11 branch of OpenShift Origin, and it doesn't reference this field.

flouthoc: @nalind That makes sense.

flouthoc:
LGTM

TomSweeneyRedHat:
LGTM
and happy green test buttons.

PR #212, opened 2022-01-31
dispatcher: Parse `--platform` via `FROM --platform=` and assign to field in stage's Builder

flouthoc:
Allows imagebuilder to read  `--platform` if specified in `FROM` and
relay it to implementations using `stage.Builder.Platform` so implementation can
actually use platform while pulling the image.

Example Demonstration/Implementation here: https://github.com/containers/buildah/pull/3757


flouthoc:
@nalind PTAL. This is the approach which i am thinking about. Following approach delegates whole `Platform` string to implementation since they are responsible for deciding about how are they going to use it.

TomSweeneyRedHat:
@flouthoc, does Docker allow for different platforms per FROM statement?  The documentation you provided wasn't clear about that.

TomSweeneyRedHat:
Would be good to have some tests added.  I think, but I'm not sure that this handles an "AS" clause.  Have you confirmed?

flouthoc:
> @flouthoc, does Docker allow for different platforms per FROM statement? The documentation you provided wasn't clear about that.

@TomSweeneyRedHat Yes I tried it last time with buildkit it supports that but only makes lot of sense when user has a `qemu-user-static` and `bindfmt_misc` setup otherwise it would not make lot of sense to user. 

But irrespective of the even legacy docker build supports `--platform` with `FROM` statements. 

> Would be good to have some tests added. I think, but I'm not sure that this handles an "AS" clause. Have you confirmed?

Yes `AS` clause is still handled it is exclusive of `--platform` and part of generic `FROM`. I'll add a test and implementation to docker client once everybody agrees on this approach.

flouthoc:
@nalind hmm i'm still thinking if we should override the variables I'll check with buildkit once if they are doing it then we can do the same ? But `go-dockerclient` for sure I'll wire it up. Just wanted to make sure if approach seemed right.

flouthoc:
@nalind I think  `--platform` does not changes or alters any of the `TARGETPLATFORM` or `BUILDPLATFORM`. I tried this using buildkit

```Dockerfile
FROM --platform=linux/arm64 alpine
ARG TARGETPLATFORM
ARG BUILDPLATFORM
RUN echo "I am running on $BUILDPLATFORM, building for $TARGETPLATFORM"
```

still shows `I am running on linux/amd64, building for linux/amd64` even though my --platform was `linux/arm64` but actual image pulled was for `linux/arm64`.

Command `sudo docker buildx build -t test .`

--platform can use `TARGETPLATFORM` or `BUILDPLATFORM` but cannot alter them.

nalind:
> @nalind I think `--platform` does not changes or alters any of the `TARGETPLATFORM` or `BUILDPLATFORM`. I tried this using buildkit
> 
> ```dockerfile
> FROM --platform=linux/arm64 alpine
> ARG TARGETPLATFORM
> ARG BUILDPLATFORM
> RUN echo "I am running on $BUILDPLATFORM, building for $TARGETPLATFORM"
> ```
> 
> still shows `I am running on linux/amd64, building for linux/amd64` even though my --platform was `linux/arm64` but actual image pulled was for `linux/arm64`.
> 
> Command `sudo docker buildx build -t test .`
> 
> --platform can use `TARGETPLATFORM` or `BUILDPLATFORM` but cannot alter them.

Apparently `docker build` doesn't set the `TARGETPLATFORM` args based on `FROM --platform`, but does set it to match a `--platform` flag passed on the command line, and there's no correspondence enforced between the two.  The output image is marked as being for `TARGETPLATFORM`.  That suggests that `TARGETPLATFORM` is more about supporting cross-compiling, and `FROM --platform` isn't really tied to it at all.  So I guess we don't have to worry about setting `TARGETPLATFORM` based on `FROM` after all.

flouthoc:
> > @nalind I think `--platform` does not changes or alters any of the `TARGETPLATFORM` or `BUILDPLATFORM`. I tried this using buildkit
> > ```dockerfile
> > FROM --platform=linux/arm64 alpine
> > ARG TARGETPLATFORM
> > ARG BUILDPLATFORM
> > RUN echo "I am running on $BUILDPLATFORM, building for $TARGETPLATFORM"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > still shows `I am running on linux/amd64, building for linux/amd64` even though my --platform was `linux/arm64` but actual image pulled was for `linux/arm64`.
> > Command `sudo docker buildx build -t test .`
> > --platform can use `TARGETPLATFORM` or `BUILDPLATFORM` but cannot alter them.
> 
> Apparently `docker build` doesn't set the `TARGETPLATFORM` args based on `FROM --platform`, but does set it to match a `--platform` flag passed on the command line, and there's no correspondence enforced between the two. The output image is marked as being for `TARGETPLATFORM`. That suggests that `TARGETPLATFORM` is more about supporting cross-compiling, and `FROM --platform` isn't really tied to it at all. So I guess we don't have to worry about setting `TARGETPLATFORM` based on `FROM` after all.

@nalind Cool then i guess just instrumenting docker-client is enough right, i'll also add a test.

dockerclient/client.go
```diff
@@ -1050,7 +1051,7 @@ func (e *ClientExecutor) archiveFromContainer(from string, src, dst string) (io.
 		containerConfig = other.Container.Config
 	} else {
 		klog.V(5).Infof("Creating a container temporarily for image input from %q in %s", from, src)
-		_, err := e.LoadImage(from)
+		_, err := e.LoadImage(from, "")
```
flouthoc, line 1054: Do we need platform here as well ? I'm trying to avoid breaking public API where its not needed.

dockerclient/client.go
```diff
@@ -556,7 +556,7 @@ func randSeq(source string, n int) (string, error) {
 
 // LoadImage checks the client for an image matching from. If not found,
 // attempts to pull the image and then tries to inspect again.
-func (e *ClientExecutor) LoadImage(from string) (*docker.Image, error) {
+func (e *ClientExecutor) LoadImage(from string, platform string) (*docker.Image, error) {
```
flouthoc, line 559: Breaking changes in Public API

nalind: This could be handled by renaming the method to `LoadImageForPlatform()`, or something along those lines, and letting it take the platform spec as you're describing, defaulting to the current platform if it's passed an empty value (as that's what happens when go-dockerclient's `PullImageOptions.Platform` isn't set, and what the engine does when the API request it receives doesn't have one) and adding a new `LoadImage()`, with the same arguments and return values as the current one, that just calls `LoadImageForPlatform()` with an empty platform value.

dockerclient/client.go
nalind: This code path is called while we're handling a COPY instruction, so unless you're also aiming to add a `Platform` field to the `Copy` type, there won't be a value to pass here.

flouthoc: ackd fixed.

dockerclient/client.go
flouthoc: Thanks. Pushed a new version.

builder.go
```diff
@@ -303,6 +303,9 @@ type Builder struct {
 	PendingCopies  []Copy
 
 	Warnings []string
+	// Raw platform string specified with `FROM --platform` of the stage
+	// Its upto the implementation or client to parse and use this field
```
nalind, line 307: Nit: typo "upto"

dispatchers_test.go
```diff
@@ -660,3 +660,27 @@ func TestDispatchRunFlags(t *testing.T) {
...
+	original := "FROM --platform=linux/arm64 busybox"
+
+	if err := from(&mybuilder, args, nil, flags, original); err != nil {
+		t.Errorf("dispatchAdd error: %v", err)
+	}
+
+	if mybuilder.Platform != expectedPlatform {
+		t.Errorf("Expected %v, to match %v\n", expectedPlatform, mybuilder.Platform)
+	}
+
```
nalind, line 685: Nit: extra blank line

nalind (commented):
A couple of formatting nits, otherwise LGTM.

flouthoc:
@rhatdan @nalind Fixed the nits. Does it look better now.

PR #213, opened 2022-01-31
dockerclient: always use force=true when removing images

nalind:
Always specify `force=true` when removing images.  Use a private method to avoid having to set up the `RemoveImageOptions` every time.


rhatdan:
LGTM
@TomSweeneyRedHat @flouthoc PTAL

flouthoc (approved):
LGTM

PR #214, opened 2022-02-01
dockerclient: tighten up cleanup logic

nalind:
When cleaning up at the end of a build, instead of iterating through stages and running their cleanup logic in whatever order a map of the stages gives them to us, run through them in the exact reverse of the order in which they were executed.

Reorder the cleanups done for each stage so that if the stage's result was committed to create an image, we try to remove that image before removing the stage's container, and if that container was based on a scratch image, we remove the scratch image after removing the container.

Removing an image produced from an intermediate stage will fail if that stage's image is an ancestor of the final image, so ignore errors we encounter while trying to remove those images.  Treat every other error we encounter while trying to remove a container or an image during conformance tests as a failure in the test.

### DIFF
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -113,7 +113,7 @@ type ClientExecutor struct {
 	LogFn func(format string, args ...interface{})
 
 	// Deferred is a list of operations that must be cleaned up at
-	// the end of execution. Use Release() to handle these.
+	// the end of execution. Use Release() to invoke all of these.
 	Deferred []func() error
 
 	// Volumes handles saving and restoring volumes after RUN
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

flouthoc (approved):
@nalind I think we invoke `WithName(` for all stages irrespective of if `COPY` instruction is added. Should we also change the function description a bit. While reading it seemed to me that we only create `childExecutor` if `COPY` was in the step.

But after reading code it seems we maintain child executor for every stage and `COPY` could use it as well.

LGTM

flouthoc:
LGTM

PR #215, opened 2022-02-02
dockerclient: COPY --from should ignore the source's WORKDIR

nalind:
When COPY is given --from=${stage|image}, the stage's or image's WORKDIR should not be incorporated into the path where we go to look for the source when the source is not specified with an absolute path name.  We previously incorporated the WORKDIR value, but that is not compatible with what `docker build` does.

### DIFF
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -11,7 +11,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"runtime"
 	"sort"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@flouthoc @umohnani8 @TomSweeneyRedHat @vrothberg PTAL

vrothberg (commented):
/lgtm

vrothberg:
LGTM

vrothberg:
I don't think I can merge here.

PR #216, opened 2022-02-02
dockerclient: copying a directory from another stage should discard perms

nalind:
Copying a directory from another stage shouldn't preserve the permissions of the directory itself, to be compatible with the behavior we see in `docker build`.


rhatdan:
LGTM
@flouthoc @TomSweeneyRedHat @umohnani8 @flouthoc PTAL

flouthoc:
Sorry I missed this yesterday. But PR LGTM

PR #218, opened 2022-02-02
dockerclient: fix case where COPY --from matches a file via a wildcard

### LINKED ISSUE #182: Multi-stage Quarkus GraalVM builds do not produce runnable images
Following the [Quarkus](https://quarkus.io/guides/building-native-image) building a native image instructions with the getting-started from the [quickstarts](https://github.com/quarkusio/quarkus-quickstarts.git) using the [multi-stage Dockerfile](https://quarkus.io/guides/building-native-image#using-a-multi-stage-docker-build) creates an unusable image.

Running the resultant image produces the following error:
```
user$ docker run -i --rm -p 8080:8080 quarkus-hello-world
docker: Error response from daemon: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "./application": permission denied: unknown.
```

nalind:
If we COPY --from content from a container using a globbing pattern, and that globbing pattern matches one non-directory, and the destination location doesn't exist yet, copy that item to the destination location's path exactly, potentially renaming it, rather than creating a directory using the destination path, and copying the item into it as a new item with the same basename.
Fixes #182.

Teach the dockerclient TestConformanceInternal test how to run multi-stage builds.

### DIFF
--- a/dockerclient/archive.go
+++ b/dockerclient/archive.go
@@ -50,7 +50,7 @@ func FilterArchive(r io.Reader, w io.Writer, fn TransformFileFunc) error {
 		var body io.Reader = tr
 		name := h.Name
 		data, ok, skip, err := fn(h, tr)
-		klog.V(6).Infof("Transform %s -> %s: data=%t ok=%t skip=%t err=%v", name, h.Name, data != nil, ok, skip, err)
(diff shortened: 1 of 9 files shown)


TomSweeneyRedHat:
LGTM
but definitely want other eyeballs on this

rhatdan:
LGTM

Issue #87, opened 2018-07-10
directory not created with WORKDIR and VOLUME


sosiouxme:
```
$ cat Dockerfile

FROM centos:7
WORKDIR /foo
VOLUME  [ "/foo" ]
RUN echo

$ imagebuilder .
--> FROM centos:7 as 0
--> WORKDIR /foo
--> VOLUME  [ "/foo" ]
--> RUN echo
API error (404): {"message":"lstat /var/lib/docker/overlay2/6b7c9551518e64992927f58da362b38949eddbb602f5eea6e15a9278be278bdb/merged/foo/: no such file or directory"}
```

You will also get a different error without the `VOLUME` line:

`/bin/sh: line 0: cd: /foo: No such file or directory`

Docker creates the dir implicitly.

The workaround is to `RUN mkdir /foo` before using it as a volume or workdir.

sosiouxme:
@smarterclayton noticed this one because OSBS adds a `RUN rm repos` line at the end of every build.

sosiouxme:
@smarterclayton just to bump this one...

sosiouxme:
> Workdir is supposed to implicitly create a dir if you have a run statement.
>

Yes... the problem is imagebuilder apparently doesn't.

smarterclayton:
A similar fix could work yes.

sosiouxme:
@smarterclayton just bumping issues in this repo... I haven't had a chance to look at them either :(

sosiouxme:
https://github.com/openshift/imagebuilder/pull/113

sosiouxme:
Fixed!

PR #219, opened 2022-03-23
builder: create WORKDIR with USER ownership

giuseppe:
let the executor deal with creating the WorkingDirectory.

Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -44,6 +44,7 @@ type Run struct {
 type Executor interface {
 	Preserve(path string) error
 	EnsureContainerPath(path string) error
+	EnsureContainerPathAs(path, user string, mode *os.FileMode) error
 	Copy(excludes []string, copies ...Copy) error
 	Run(run Run, config docker.Config) error
 	UnrecognizedInstruction(step *Step) error
@@ -58,6 +59,15 @@ func (logExecutor) Preserve(path string) error {
 
 func (logExecutor) EnsureContainerPath(path string) error {
 	log.Printf("ENSURE %s", path)
+	return nil
+}
+
+func (logExecutor) EnsureContainerPathAs(path, user string, mode *os.FileMode) error {
+	if mode != nil {
+		log.Printf("ENSURE %s AS %q with MODE=%q", path, user, *mode)
+	} else {
+		log.Printf("ENSURE %s AS %q", path, user)
+	}
 	return nil
 }
 
@@ -85,6 +95,10 @@ func (noopExecutor) Preserve(path string) error {
 }
 
 func (noopExecutor) EnsureContainerPath(path string) error {
+	return nil
+}
+
+func (noopExecutor) EnsureContainerPathAs(path, user string, mode *os.FileMode) error {
 	return nil
 }
 
--- a/builder.go
+++ b/builder.go
@@ -392,7 +392,7 @@ func (b *Builder) Run(step *Step, exec Executor, noRunsRemaining bool) error {
 	}
 
 	if len(b.RunConfig.WorkingDir) > 0 {
-		if err := exec.EnsureContainerPath(b.RunConfig.WorkingDir); err != nil {
+		if err := exec.EnsureContainerPathAs(b.RunConfig.WorkingDir, b.RunConfig.User, nil); err != nil {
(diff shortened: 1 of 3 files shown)


giuseppe:
I've tried the reproducer here https://github.com/openshift/imagebuilder/issues/87 and could not reproduce the issue.

Needed for: https://github.com/containers/buildah/pull/3840

rhatdan:
LGTM, but we need @nalind opinion on this.

nalind:
Is there an explanation of why this change is being made?  Conformance tests don't get run in CI; do they continue to pass with this change?

giuseppe:
> Do I need Docker to run the conformance tests? I've pointed it to Podman (running locally with `sudo podman system service -t 0 unix:///var/run/docker.sock`) and I get all sort of failures.

stupid question.  We clearly need Docker, going to test with that.

nalind:
> if we postpone the mkdir to [containers/buildah#3840](https://github.com/containers/buildah/pull/3840) then we can create it with the correct ownership.

Please make sure that's noted in the commit log.

builder.go
```diff
@@ -61,6 +62,15 @@ func (logExecutor) EnsureContainerPath(path string) error {
 	return nil
 }
 
+func (logExecutor) EnsureContainerPathAs(path, user string, mode *os.FileMode) error {
+	if mode != nil {
+		log.Printf("ENSURE %s AS %q with MODE=%q", path, user, *mode)
```
rhatdan, line 67: Is this debug code?

nalind: The `logExecutor` is used by tests.

giuseppe: no, this is the logExecutor that just logs what is happening

rhatdan:
@giuseppe could you force push this to trigger CI.

flouthoc (approved):
LGTM @nalind PTAL

dockerclient/client.go
```diff
@@ -685,6 +685,11 @@ func (e *ClientExecutor) EnsureContainerPath(path string) error {
 	return e.createOrReplaceContainerPathWithOwner(path, 0, 0)
 }
 
+func (e *ClientExecutor) EnsureContainerPathAs(path, user string, mode *os.FileMode) error {
+	// FIXME: user and mode are currently ignored.
+	return e.createOrReplaceContainerPathWithOwner(path, 0, 0)
```
nalind, line 690: Must not be merged before this is resolved.

giuseppe: I've not implemented it because we would have to convert the userspec into UID:GID.

Do you have any suggestions?  Do we need to copy the `Builder.userForCopy` function from buildah?

nalind: You might refactor parts of the dockerclient `CopyContainer()` method, which has to do similar work to support a `--chown` flag.

dockerclient/client.go
```diff
@@ -682,10 +682,27 @@ func (e *ClientExecutor) Preserve(path string) error {
...
+func (e *ClientExecutor) EnsureContainerPathAs(path, user string, mode *os.FileMode) error {
+	uid, gid := 0, 0
+
+	u, g, err := e.getUser(user)
+	if err == nil {
+		uid = u
+		gid = g
+	}
+
+	// FIXME: user and mode are currently ignored.
```
rhatdan, line 697: Is this FIXME still correct?

giuseppe: I forgot to drop it.

Cleaned up now

TomSweeneyRedHat:
LGTM

giuseppe:
fixed, Buildah tests are also green: https://github.com/containers/buildah/pull/3840

nalind:
Did we run the conformance tests?

giuseppe:
I've used this branch in https://github.com/containers/buildah/pull/3840

nalind:
I was referring to this repository's conformance tests, which CI doesn't run.

nalind:
Something like https://github.com/nalind/imagebuilder/commit/b2bca3bd0185327124628ebc0aee0dde000dcc45.

PR #222, opened 2022-03-29
conformance: test non-root USER before WORKDIR

nalind:
Add a test that checks who owns a WORKDIR if it's created after USER sets a non-root default runtime user.  Try to run conformance tests in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 
 go:
-  - "1.16"
   - "1.17"
+  - "1.18"
 
 install:
 
--- a/dockerclient/conformance_test.go
+++ b/dockerclient/conformance_test.go
@@ -458,6 +458,16 @@ func TestConformanceInternal(t *testing.T) {
 			ContextDir: "testdata/multistage",
 			Dockerfile: "Dockerfile.env",
 		},
+		{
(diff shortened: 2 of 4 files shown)


giuseppe:
LGTM

I've opened a similar PR to run the conformance tests using github actions: https://github.com/openshift/imagebuilder/pull/223

From a test run they seem to pass: https://github.com/giuseppe/imagebuilder/actions/runs/2059604397

giuseppe (approved):

TomSweeneyRedHat:
LGTM
and happy green tests

nalind:
Yeah, we made a change that we didn't test for conformance at the time, and it breaks unless the daemon's using BuildKit (i.e., on the traditional `docker build` path).  Options include backing out that change, or updating the conformance tests to ask for BuildKit.

rhatdan:
I believe the best solution is to move onto BuildKit conformance.  Since this is the direction that Docker is taking.

nalind:
Dumb question time: am I the only person who doesn't see results of the `test` action somewhere in this PR or in this repository?  The bot told me the tests passed about five minutes after I last re-pushed the branch, which is an unusually quick turnaround for the conformance tests.

PR #223, opened 2022-03-29
.github: run conformance tests

giuseppe:
Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,24 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  Test:
+    runs-on: ubuntu-20.04
+    name: Conformance tests
+
+    steps:
+      - uses: actions/checkout@v2.1.0
+
+      - name: Install dependencies
+        run: |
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@nalind @flouthoc @TomSweeneyRedHat PTAL

flouthoc (approved):
LGTM

Do we also need to enable github-actions on the repo

.github/workflows/test.yaml
```diff
@@ -0,0 +1,24 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  Test:
+    runs-on: ubuntu-20.04
```
flouthoc, line 7: I think conformance test also needs docker daemon to be running. Is it available on `ubuntu-20.04` by default.

giuseppe:
is there any interest in this PR or should I close it?

flouthoc:
@giuseppe I think we should have it in upstream cause sometimes its pain to run conformance test locally. @nalind @rhatdan PTAL

PR #224, opened 2022-04-04
dispatcher: honor default builtin-args while processing `FROM --platform=`

flouthoc:
While processing inline `--platform=` in `FROM` , imagebuilder ignores
default builtinArgs so for example if user wants to use inbuilt default arg
`FROM --platform=$BUILDPLATFORM` it should automatically replace it with
default value of BUILDPLATFORM.

Fixes
```Dockerfile
FROM --platform=$BUILDPLATFORM busybox
RUN echo "do something"
```

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -234,7 +234,9 @@ func from(b *Builder, args []string, attributes map[string]bool, flagArgs []stri
 			return fmt.Errorf("Windows does not support FROM scratch")
 		}
 	}
+	defaultArgs := envMapAsSlice(builtinBuildArgs)
 	userArgs := mergeEnv(envMapAsSlice(b.Args), b.Env)
+	userArgs = mergeEnv(defaultArgs, userArgs)
 	for _, a := range flagArgs {
 		arg, err := ProcessWord(a, userArgs)
 		if err != nil {
--- a/dispatchers_test.go
+++ b/dispatchers_test.go
@@ -683,3 +683,26 @@ func TestDispatchFromFlags(t *testing.T) {
 		t.Errorf("Expected %v, to match %v\n", expectedPlatform, mybuilder.Platform)
 	}
 }
+
+func TestDispatchFromFlagsAndUseBuiltInArgs(t *testing.T) {
+	expectedPlatform := localspec.OS + "/" + localspec.Architecture
+	mybuilder := Builder{
+		RunConfig: docker.Config{
+			WorkingDir: "/root",
+			Cmd:        []string{"/bin/sh"},
+			Image:      "busybox",
+		},
+	}
+
+	flags := []string{"--platform=$BUILDPLATFORM"}
+	args := []string{""}
+	original := "FROM --platform=$BUILDPLATFORM busybox"
+
+	if err := from(&mybuilder, args, nil, flags, original); err != nil {
+		t.Errorf("dispatchAdd error: %v", err)
+	}
+
+	if mybuilder.Platform != expectedPlatform {
+		t.Errorf("Expected %v, to match %v\n", expectedPlatform, mybuilder.Platform)
+	}
+}


flouthoc:

dispatchers.go
```diff
@@ -234,7 +234,9 @@ func from(b *Builder, args []string, attributes map[string]bool, flagArgs []stri
 			return fmt.Errorf("Windows does not support FROM scratch")
 		}
 	}
+	defaultArgs := envMapAsSlice(builtinBuildArgs)
```
nalind, line 237: Are the default values all correct when `--platform` is specified differently for different stages?

flouthoc: @nalind mostly we only use `$BUILDPLATFORM` or `$TARGETPLATFORM` in inline platform syntax so yes default values are correct for different stages. However there is still a small issue with the way buildah populates `$TARGETPLATFORM` but its a small fix on buildah side not on imagebuilder side.

flouthoc: Created a PR in buildah which exercises this and fixes `$TARGETPLATFORM` issue: https://github.com/containers/buildah/pull/3899

nalind: What about the `dockerclient` client?

TomSweeneyRedHat:
LGTM
if @nalind is happy with this.

flouthoc:
@nalind Could PTAL again.

dispatchers.go
nalind: The behavior not being in `docker build` means that we can't write a conformance test for it, and that is very unfortunate, but that doesn't tell me whether someone who uses the `imagebuilder` CLI will get a surprising result if their Dockerfile tries to reference these arguments.

A different change we previously made at the parsing level for `buildah`'s sake caused `imagebuilder` to switch from reporting an error when a specific option was used to silently ignoring it, and that tripped up our colleagues who use `imagebuilder` when they found it no longer complained when they tried to use the option.  IIRC they got pretty far down the road before noticing it, since ignoring that option still produced the correct result some of the time.  I'm trying to avoid repeating that mistake.

flouthoc: @nalind hmm that is true do you have any suggestion about what should we do for `dockerclient` for such cases ? Would a warning work here.

nalind: It would need to either flag the attempt to use the unsupported argument as an error, or expand it correctly.  What happens with the current patch?

flouthoc: > @nalind For `dockerclient` we are already setting args correctly so `$BUILDPLATFORM` will be set correctly but other options like `$TARGETPLATFORM` will be only exercised by docker deamon if `"features": { "buildkit": true }` is enabled on the server side. So if docker daemon is running without buildkit certain things will be `no-op` but I think imagebuilder can not help a lot in such cases.

Its pretty much like above everything works fine except its just that `$TARGETPLATFORM` might be `no-op` depending upon the fact if `dockerserver` supports buildkit or not but I also think this part might be out of the scope of this PR itself.

But PR only talks about implementation of `FROM --platform=` which works fine for `dockerclient` and we added implementation for dockerclient here: https://github.com/openshift/imagebuilder/pull/212

nalind: The docker daemon doesn't evaluate the Dockerfile, we do.

flouthoc: > The docker daemon doesn't evaluate the Dockerfile, we do.

Oh my bad i mixed things, if we use `TARGETPLATFORM` as `-build-arg` with `imagebuilder` it will work but we can't expect `$TARGETPLATFORM` to get correctly populated cause we don't have cli arg like `buildah build --platform` or `docker build --platform` in imagebuilder to correctly set the value of `TARGETPLATFORM` correct me if i got this wrong.

So its just that with `imagebuilder` it is expected from user to set the value of `TARGETGETPLATFORM` using `-build-arg` otherwise there is no way set the correct value of `TARGETPLATFORM` and default value is always https://github.com/openshift/imagebuilder/blob/master/dispatchers.go#L35

I have a feeling that this is just expected behavior and I'm overthinking this.

rhatdan: Wouldn't TAGETGETPLATFORM just default to the local platform then?

rhatdan:
/approve
LGTM

flouthoc:
@nalind @rhatdan Performed conformance test on this PR with `buildkit` enabled.

with and without below diff I got same output as: https://pastebin.com/raw/mrb63TWE 
```diff
diff --git a/dockerclient/conformance_test.go b/dockerclient/conformance_test.go
index 66508f3..5585c9c 100644
--- a/dockerclient/conformance_test.go
+++ b/dockerclient/conformance_test.go
@@ -142,7 +142,7 @@ func TestCopyFrom(t *testing.T) {
                        e.Out, e.ErrOut = out, out
                        b := imagebuilder.NewBuilder(nil)
                        dockerfile := fmt.Sprintf(`
-                               FROM busybox AS base
+                               FROM --platform=$BUILDPLATFORM busybox AS base
                                RUN %s
                                FROM busybox
                                %s
```

Two tests always fail for me even on the `master` branch without this PR and with or without `buildkit` enabled ( also both tests seems unrelated to this PR )
* TestMount 
* TestConformanceInternal/novolume

nalind:
A `chmod -R go-w ./dockerclient/testdata` is generally needed before running the conformance tests, as they make assumptions about permissions of files in the source tree which are not always correct.

flouthoc:
Thanks @nalind after `chmod -R go-w ./dockerclient/testdata`  `TestMount` passes now just `TestConformanceInternal/novolume` fails for me both on this PR and `master` branch as well.

Output: https://pastebin.com/raw/HVGGGAya

Independent run for ``TestConformanceInternal/novolume``

```console
go test -v -tags conformance -timeout 3h ./dockerclient -run=TestConformanceInternal/novolume
=== RUN   TestConformanceInternal
=== RUN   TestConformanceInternal/novolume
    conformance_test.go:1068: conformance-dockerbuild-28-docker-0-5 and conformance-dockerbuild-28-direct-0-5 differ at var/lib/not-in-this-image/:
        &tar.Header{Typeflag:0x35, Name:"var/lib/not-in-this-image/", Linkname:"", Size:0, Mode:17389, Uid:0, Gid:0, Uname:"", Gname:"", ModTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), AccessTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), ChangeTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Devmajor:0, Devminor:0, Xattrs:map[string]string(nil), PAXRecords:map[string]string(nil), Format:2}
        &tar.Header{Typeflag:0x35, Name:"var/lib/not-in-this-image/", Linkname:"", Size:0, Mode:16877, Uid:0, Gid:0, Uname:"", Gname:"", ModTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), AccessTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), ChangeTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Devmajor:0, Devminor:0, Xattrs:map[string]string(nil), PAXRecords:map[string]string(nil), Format:2}
    conformance_test.go:1068: conformance-dockerbuild-28-docker-0-5 and conformance-dockerbuild-28-direct-0-5 differ at var/lib/:
        &tar.Header{Typeflag:0x35, Name:"var/lib/", Linkname:"", Size:0, Mode:17389, Uid:0, Gid:0, Uname:"", Gname:"", ModTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), AccessTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), ChangeTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Devmajor:0, Devminor:0, Xattrs:map[string]string(nil), PAXRecords:map[string]string(nil), Format:2}
        &tar.Header{Typeflag:0x35, Name:"var/lib/", Linkname:"", Size:0, Mode:16877, Uid:0, Gid:0, Uname:"", Gname:"", ModTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), AccessTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), ChangeTime:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Devmajor:0, Devminor:0, Xattrs:map[string]string(nil), PAXRecords:map[string]string(nil), Format:2}
    conformance_test.go:1083: a(conformance-dockerbuild-28-docker-0-5)=map[] b(conformance-dockerbuild-28-direct-0-5)=map[] diff=map[var/lib/:[<0 dtrwxr-xr-x> <0 drwxr-xr-x>] var/lib/not-in-this-image/:[<0 dtrwxr-xr-x> <0 drwxr-xr-x>]]
    conformance_test.go:946: 28: full Docker build was not equivalent to squashed image metadata /tmp/dockerbuild-conformance-1936807101/Dockerfile
=== RUN   TestConformanceInternal/novolumenorun
--- FAIL: TestConformanceInternal (38.38s)
    --- FAIL: TestConformanceInternal/novolume (23.33s)
    --- PASS: TestConformanceInternal/novolumenorun (15.05s)
FAIL
FAIL	github.com/openshift/imagebuilder/dockerclient	38.386s
FAIL
make: *** [Makefile:10: test-conformance] Error 1
```

nalind:
That's... is docker setting the sticky bit on those directories?  Is this a difference in behavior between docker-with-buildkit and docker-without-buildkit?

flouthoc:
Was failing for me irrespective of the buildkit but i can confirm again.

flouthoc:
It was my docker version updating it fixed for both buildkit and regular docker.

flouthoc:
Passes now 
```console
$ sudo go test -v -tags conformance -timeout 3h ./dockerclient -run=TestConformanceInternal/novolume
=== RUN   TestConformanceInternal
=== RUN   TestConformanceInternal/novolume
=== RUN   TestConformanceInternal/novolumenorun
--- PASS: TestConformanceInternal (9.63s)
    --- PASS: TestConformanceInternal/novolume (6.67s)
    --- PASS: TestConformanceInternal/novolumenorun (2.96s)
PASS
ok  	github.com/openshift/imagebuilder/dockerclient	9.630s
```

siretart:
This patch introduces tests that fail on arm64, armhf and armel with:

On arm64:
```
=== RUN   TestDispatchFromFlagsAndUseBuiltInArgs
    dispatchers_test.go:798: Expected linux/arm64, to match linux/arm64/v8
--- FAIL: TestDispatchFromFlagsAndUseBuiltInArgs (0.00s)
```

on armel:
```
=== RUN   TestDispatchFromFlagsAndUseBuiltInArgs
    dispatchers_test.go:798: Expected linux/arm, to match linux/arm/v8
```

On armhf:

```
=== RUN   TestDispatchFromFlagsAndUseBuiltInArgs
    dispatchers_test.go:798: Expected linux/arm, to match linux/arm/v7
--- FAIL: TestDispatchFromFlagsAndUseBuiltInArgs (0.00s)
```

PR #225, opened 2022-04-19
healthcheck: set appropriate defaults for `interval`, `timeout` and `retries`

flouthoc:
Set appropriate defaults for `--interval`, `--timeout` and `--retries` when
processing a Containerfile with build format as docker.

See: https://docs.docker.com/engine/reference/builder/#healthcheck
Closes: https://github.com/containers/podman/issues/13912

### DIFF
--- a/builder_test.go
+++ b/builder_test.go
@@ -631,6 +631,21 @@ func TestBuilder(t *testing.T) {
 			},
 		},
 		{
+			Dockerfile: "dockerclient/testdata/Dockerfile.healthcheckdefaults",
+			From:       "debian",
+			Config: docker.Config{
+				Image: "debian",
+				Cmd:   []string{"/bin/sh", "-c", "/app/main.sh"},
+				Healthcheck: &docker.HealthConfig{
+					StartPeriod: 0 * time.Second,
+					Interval:    30 * time.Second,
+					Timeout:     30 * time.Second,
+					Retries:     3,
+					Test:        []string{"CMD-SHELL", "/app/check.sh --quiet"},
+				},
+			},
+		},
+		{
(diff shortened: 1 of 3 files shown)


flouthoc:

rhatdan:
/approve
LGTM
@nalind @giuseppe PTAL

builder_test.go
```diff
@@ -630,6 +630,21 @@ func TestBuilder(t *testing.T) {
...
+				Cmd:   []string{"/bin/sh", "-c", "/app/main.sh"},
+				Healthcheck: &docker.HealthConfig{
+					StartPeriod: 0 * time.Second,
+					Interval:    30 * time.Second,
+					Timeout:     30 * time.Second,
+					Retries:     3,
+					Test:        []string{"CMD-SHELL", "/app/check.sh --quiet"},
+				},
+			},
+		},
```
TomSweeneyRedHat, line 647: I'm probably overthinking, but I just wonder if we're starting to brush up against DockerHub rate limits.  No need to adjust,  just thought I'd put that thought out there.

TomSweeneyRedHat:
LGTM, once @nalind 's concerns are addressed.

builder_test.go
nalind: We tend to use one of a handful of base images for various tests, and the docker daemon hangs on to them between tests, but you're not wrong to be concerned that it will become a problem at some point.

flouthoc:
It is documented here https://docs.docker.com/engine/reference/builder/#healthcheck that `--interval` is a dockerfile default but looks like docker injects these defaults while creating the container and not in image. So this PR is replaced by https://github.com/containers/podman/pull/13928

I have also verified this against conformance tests so this PR must not go in and alt is: https://github.com/containers/podman/pull/13928

PR #227, opened 2022-05-17
Add option `--keep-ownership` to ADD/COPY

### LINKED ISSUE #226: [Feature] Add --keep-ownership option to add/copy commands to keep file ownership.
Add an option to `ADD/COPY` to keep the ownership of the file also in the container.

Ref: https://github.com/containers/buildah/pull/3767

Dexus:
This PR is to keep the ownership of local files, when used with ADD or COPY - BUT WITHOUT `--chown` flag

fixes #226 , Ref: https://github.com/containers/buildah/pull/3767,  containers/buildah/pull/4001

What are the the intention for the PR:

ADD or COPY 
- without `--chown` and `--keep-ownership` result in `root:root` for the files.
- with `--chown 123:456` result in `123:456` for the files.
- with `--keep-ownership` result in `UID:GID` for the files, same as the origin source.

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -140,6 +140,7 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 	}
 	var chown string
 	var chmod string
+	var preserveOwnership bool
 	last := len(args) - 1
 	dest := makeAbsolute(args[last], b.RunConfig.WorkingDir)
 	userArgs := mergeEnv(envMapAsSlice(b.Args), b.Env)
@@ -149,6 +150,8 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 			return err
 		}
 		switch {
+		case arg == "--keep-ownership":
+			preserveOwnership = true
 		case strings.HasPrefix(arg, "--chown="):
 			chown = strings.TrimPrefix(arg, "--chown=")
 		case strings.HasPrefix(arg, "--chmod="):
@@ -158,10 +161,15 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 				return err
 			}
 		default:
-			return fmt.Errorf("ADD only supports the --chmod=<permissions> and the --chown=<uid:gid> flag")
+			return fmt.Errorf("ADD only supports the --chmod=<permissions>, --chown=<uid:gid>, and the --keep-ownership flag")
 		}
 	}
-	b.PendingCopies = append(b.PendingCopies, Copy{Src: args[0:last], Dest: dest, Download: true, Chown: chown, Chmod: chmod})
+
+	if chown != "" && preserveOwnership {
+		return errConflictChownKeepOwnerShip("ADD")
+	}
+
+	b.PendingCopies = append(b.PendingCopies, Copy{Src: args[0:last], Dest: dest, Download: true, Chown: chown, Chmod: chmod, PreserveOwnership: preserveOwnership})
 	return nil
 }
 
@@ -177,6 +185,7 @@ func dispatchCopy(b *Builder, args []string, attributes map[string]bool, flagArg
 	dest := makeAbsolute(args[last], b.RunConfig.WorkingDir)
 	var chown string
 	var chmod string
+	var preserveOwnership bool
 	var from string
 	userArgs := mergeEnv(envMapAsSlice(b.Args), b.Env)
 	for _, a := range flagArgs {
@@ -185,6 +194,8 @@ func dispatchCopy(b *Builder, args []string, attributes map[string]bool, flagArg
 			return err
 		}
 		switch {
+		case arg == "--keep-ownership":
+			preserveOwnership = true
 		case strings.HasPrefix(arg, "--chown="):
 			chown = strings.TrimPrefix(arg, "--chown=")
 		case strings.HasPrefix(arg, "--chmod="):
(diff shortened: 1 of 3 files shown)


dispatchers.go
```diff
@@ -158,10 +161,15 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 				return err
 			}
 		default:
-			return fmt.Errorf("ADD only supports the --chmod=<permissions> and the --chown=<uid:gid> flag")
+			return fmt.Errorf("ADD only supports the --keep-ownership --chmod=<permissions> and the --chown=<uid:gid> flag")
 		}
 	}
-	b.PendingCopies = append(b.PendingCopies, Copy{Src: args[0:last], Dest: dest, Download: true, Chown: chown, Chmod: chmod})
+
+	if chown != "" && preserveOwnership == true {
```
flouthoc, line 168:
```suggestion
	if chown != "" && preserveOwnership {
```

dispatchers.go
```diff
@@ -158,10 +161,15 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 				return err
 			}
 		default:
-			return fmt.Errorf("ADD only supports the --chmod=<permissions> and the --chown=<uid:gid> flag")
+			return fmt.Errorf("ADD only supports the --keep-ownership --chmod=<permissions> and the --chown=<uid:gid> flag")
```
flouthoc, line 164:
```suggestion
			return fmt.Errorf("ADD only supports the --keep-ownership, --chmod=<permissions> and the --chown=<uid:gid> flag")
```

dispatchers.go
```diff
@@ -196,10 +207,15 @@ func dispatchCopy(b *Builder, args []string, attributes map[string]bool, flagArg
 		case strings.HasPrefix(arg, "--from="):
 			from = strings.TrimPrefix(arg, "--from=")
 		default:
-			return fmt.Errorf("COPY only supports the --chmod=<permissions> --chown=<uid:gid> and the --from=<image|stage> flags")
+			return fmt.Errorf("COPY only supports the --keep-ownership --chmod=<permissions> --chown=<uid:gid> and the --from=<image|stage> flags")
```
flouthoc, line 210:
```suggestion
			return fmt.Errorf("COPY only supports the --keep-ownership, --chmod=<permissions> --chown=<uid:gid> and the --from=<image|stage> flags")
```

dispatchers.go
```diff
@@ -196,10 +207,15 @@ func dispatchCopy(b *Builder, args []string, attributes map[string]bool, flagArg
...
 		}
 	}
-	b.PendingCopies = append(b.PendingCopies, Copy{From: from, Src: args[0:last], Dest: dest, Download: false, Chown: chown, Chmod: chmod})
+
+	if chown != "" && preserveOwnership == true {
```
flouthoc, line 214:
```suggestion
	if chown != "" && preserveOwnership {
```

flouthoc:
I think this flag would be buildah only but it will also get parsed for `dockerclient` as well but would not be recognized by docker daemon, maybe we should just drop this `fmt.Errorf` from imagebuilder and let implementations take care of it. @nalind WDYT ?

```go
default:
   return fmt.Errorf("ADD only supports the --keep-ownership --chmod=<permissions> and the --chown=<uid:gid> flag")
			
```

flouthoc:
@Dexus Could you squash commits into one I don't think additional commits are needed you can keep the only first original commit. Adding a small test for this should be easy in `dispatcher_test`.

flouthoc:
> I think this flag would be buildah only but it will also get parsed for `dockerclient` as well but would not be recognized by docker daemon, maybe we should just drop this `fmt.Errorf` from imagebuilder and let implementations take care of it. @nalind WDYT ?
> 
> ```go
> default:
>    return fmt.Errorf("ADD only supports the --keep-ownership --chmod=<permissions> and the --chown=<uid:gid> flag")
> 			
> ```

Just waiting for a review by other maintainers on this point. Otherwise LGTM ( just need to add a small test in `dispatcher_test.go` )

Dexus:
@cc @TomSweeneyRedHat, @mrunalp, @nalind, @rhatdan, @smarterclayton
Would not like to push you, but would be very happy about a quick review and approval.
Thanks, Josef

dispatchers.go
```diff
@@ -196,10 +207,15 @@ func dispatchCopy(b *Builder, args []string, attributes map[string]bool, flagArg
 		case strings.HasPrefix(arg, "--from="):
 			from = strings.TrimPrefix(arg, "--from=")
 		default:
-			return fmt.Errorf("COPY only supports the --chmod=<permissions> --chown=<uid:gid> and the --from=<image|stage> flags")
+			return fmt.Errorf("COPY only supports the --chmod=<permissions>, --chown=<uid:gid>, --from=<image|stage>, and the --keep-ownership flags")
 		}
 	}
-	b.PendingCopies = append(b.PendingCopies, Copy{From: from, Src: args[0:last], Dest: dest, Download: false, Chown: chown, Chmod: chmod})
+
+	if chown != "" && preserveOwnership {
```
rhatdan, line 214: Wy doesn't this return an error?
Shoulding --chown=foobar --keep-ownership conflict?

rhatdan:
What are the defaults right now.

Default -> Ownership is root or ownership is the current User?
--chown USER -> Ownership is USER
--keep-ownership -> Current Owner UID of source?

Dexus:
> What are the defaults right now.
> 
> Default -> Ownership is root or ownership is the current User? 

Default is currently that the file is is `root:root`

> --chown USER -> Ownership is USER 
> --keep-ownership -> Current Owner UID of source?
`--keep-ownership` keeps the same UID:GID of the source.

dispatchers.go
Dexus: I did change it, now an error is returned.

Dexus:
> > I think this flag would be buildah only but it will also get parsed for `dockerclient` as well but would not be recognized by docker daemon, maybe we should just drop this `fmt.Errorf` from imagebuilder and let implementations take care of it. @nalind WDYT ?
> 
> Yes, `dockerclient` would either need to implement it or flag it as an error, to avoid a problem where it just starts silently ignoring a new flag that it doesn't implement, one which it previously flagged as an error when people tried to use it.

@nalind @flouthoc if we speak from `dockerclient` we talking not from the directory in this repo right? You mean github.com/fsouza/go-dockerclient - is it correct?

nalind:
> @nalind @flouthoc if we speak from `dockerclient` we talking ~not~ from the directory in this repo right? ~You mean github.com/fsouza/go-dockerclient - is it correct?~ I'm maybe not right up, but I'm not see where I should add it for `dockerclient`

Sorry for being unintentionally vague there - by `dockerclient` I mean the `dockerclient` package in this repository, which provides an executor for the builder logic by calling into `go-dockerclient`.  That separation, between the parser, the builder logic that drives the build, and the executor that it uses, made imagebuilder a natural choice for adding Dockerfile support to buildah.

nalind:
This doesn't seem to be an option that `docker build` supports in its current iteration, and we've historically shied away from implementing syntax features that `docker build` doesn't implement because we don't want to "fork" the format.  Has this been proposed there?

Dexus:
> This doesn't seem to be an option that `docker build` supports in its current iteration, and we've historically shied away from implementing syntax features that `docker build` doesn't implement because we don't want to "fork" the format. Has this been proposed there?

No, that was not proposed there. But I will do this. Not sure where exactly is the right place for it. Moby?

rhatdan:
Yes Moby.  But if this is a change to Dockerfile/Containerfile format, then they will probably send you to buildkit.

PR #229, opened 2022-05-26
dispatcher: consider `builtinargs` and `userargs` while processing `baseImage` name

flouthoc:
We already consider `builtinargs` and `userArgs` while computing `--from=`
flags so use same logic for computing name of baseImage while evaluating
`FROM` statements.

Allows use cases like
```Dockerfile
FROM --platform=linux/amd64 alpine as platform-amd64
# do something

FROM --platform=linux/arm64 alpine as platform-arm64
# do something

FROM platform-${TARGETARCH}
```

also allows `FROM ${SOME_DEFAULT_OR_USER_ARG}` in general.

Closes: https://github.com/containers/podman/issues/14375

### DIFF
--- a/builder_test.go
+++ b/builder_test.go
@@ -16,6 +16,7 @@ import (
 	docker "github.com/fsouza/go-dockerclient"
 
 	"github.com/openshift/imagebuilder/dockerfile/parser"
+	"github.com/containerd/containerd/platforms"
 )
 
 func TestVolumeSet(t *testing.T) {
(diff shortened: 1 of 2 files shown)


flouthoc:

rhatdan:
LGTM
Although this does not look like tests ran.

builder_test.go
```diff
@@ -209,6 +210,38 @@ RUN echo $FOO $BAR`))
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			if from != tc.expectedFrom {
+				t.Fatalf("expected %s, got %s", tc.expectedFrom, from)
+			}
+		})
+	}
+}
+
```
TomSweeneyRedHat, line 244: Do we have a test that also covers this?

`also allows FROM ${SOME_DEFAULT_OR_USER_ARG} in general.`

flouthoc: @TomSweeneyRedHat I think test in this PR should cover that as well since it performs test for `$TARGETARCH` which is one of the DEFAULT ARGS but i can add another test for user args as well.

builder_test.go
```diff
@@ -209,6 +210,38 @@ RUN echo $FOO $BAR`))
...
+		expectedFrom string
+	}{
+		{name: "override", args: map[string]string{"FOO": "bar"}, expectedFrom: "platform-"+localspec.Architecture},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			n, err := ParseDockerfile(strings.NewReader(`ARG FOO=latest
+ARG BAR=baz
+FROM platform-${TARGETARCH}
+ARG BAZ=banana
+RUN echo $FOO $BAR`))
```
nalind, line 229: This can probably be simplified from the version in the `TestHeadingArg` test, since the values for FOO, BAR, and BAZ aren't used on the FROM line.

nalind:
Code change looks fine, but the new test could use a few more entries in its slice of test cases.

builder_test.go
flouthoc: Yes we don't need extra lines only lines in the format of `FROM ${ARG}` is enough for this test.

builder_test.go
flouthoc: @TomSweeneyRedHat Added additional test case.

nalind:
LGTM

Issue #231, opened 2022-06-24
Imagebuilder expands `ARGS` which are not declared


flouthoc:
As of now imagebuilder expands args for certain instructions `ADD`, `RUN` and `FROM` which are not declared in Containerfile.

Example
```Dockerfile
FROM alpine
ADD --chown=${CUSTOM_USER} somefile .
RUN echo hello
```

```console
$ touch somefile
buildah build --build-arg CUSTOM_USER=hello -t test .
```

Build fails since it tries to use the resolved user however it should not resolve args which are not declared.

flouthoc:
@nalind If i understand the question correctly do you mean in what cases we expand args without declaration args in case of `ADD` and `RUN`.

In `ADD` we expand args for flags like `--chown=${SOME_ARG}`  and for `RUN` we expand args in `--mount` flag so something like `--mount=type=${SOME_ARG}` would be expand even if `SOME_ARG` is not declared in Containerfile

PR #232, opened 2022-06-27
dispatcher, builder: filter `args` against `allowedArgs` while processing `FROM`, `ADD` and `RUN`

flouthoc:
Following PR adds two commits to make sure we expand variables while processing `FROM`, `ADD` and `RUN` only if they are declared. See individual commits for detail.

* FROM: only resolve arg in `FROM` if declared anywhere
     Resolution of `FROM ${SOMEARG}` should happen only if `SOMEARG` was
     declared in the Containerfile.
     
 * dispatcher: filter args against allowed args while processing flags for `ADD` and `RUN`
 
 Closes: https://github.com/openshift/imagebuilder/issues/231

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -215,6 +215,17 @@ type Stage struct {
 
 func NewStages(node *parser.Node, b *Builder) (Stages, error) {
 	var stages Stages
+	var allDeclaredArgs []string
+	for _, root := range SplitBy(node, command.Arg) {
+		argNode := root.Children[0]
+		if argNode.Value == command.Arg {
+			// extract declared variable
+			s := strings.SplitN(argNode.Original, " ", 2)
+			if len(s) == 2 && (strings.ToLower(s[0]) == command.Arg) {
+				allDeclaredArgs = append(allDeclaredArgs, s[1])
+			}
+		}
+	}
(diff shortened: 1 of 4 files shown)


flouthoc:
@nalind PTAL

builder.go
```diff
@@ -215,6 +215,17 @@ type Stage struct {
 
 func NewStages(node *parser.Node, b *Builder) (Stages, error) {
 	var stages Stages
+	var allDeclaredArgs []string
+	for _, root := range SplitBy(node, command.Arg) {
+		argNode := root.Children[0]
+		if argNode.Value == command.Arg {
+			// extract declared variable
+			s := strings.SplitN(argNode.Original, " ", 2)
+			if len(s) == 2 && (strings.ToLower(s[0]) == strings.ToLower("ARG")) {
```
rhatdan, line 224: strings.ToLower("ARG") versus "arg"?

flouthoc: Yes, switched this to an already defined constant here https://github.com/openshift/imagebuilder/blob/master/dockerfile/command/command.go#L7

TomSweeneyRedHat:
LGTM
@nalind PTAL

PR #233, opened 2022-06-27
dockerclient: fix COPYing an archive to a new location

### LINKED ISSUE #230: COPY tarball treats destination as directory
Given the following Dockerfile, where `plugins.tar` is either a plain or compressed tarball which unpacks to `plugins/`:

```Dockerfile
FROM registry.access.redhat.com/ubi8

COPY plugins.tar /plugins.tar

RUN tar xf /plugins.tar && ls -l /plugins*
```

With imagebuilder 1.2.2, as with docker and podman/buildah, this works as expected.  However, with imagebuilder 1.2.3, `/plugins.tar` is created as a directory, and the file is copied therein, resulting in `/plugins.tar/plugins.tar`, and hence the subsequent `tar xf` fails with the error `Cannot read: Is a directory`.

nalind:
When COPYing an archive to a location which doesn't exist, don't force the destination to be treated as a directory that we might have to create.  We have to do that for ADD in order to be able to extract the archive's contents, but not for COPY.

Document that the `Decompress` field from `dockerclient.CopyInfo` is both never set and ignored.

Should fix #230.

### DIFF
--- a/dockerclient/archive.go
+++ b/dockerclient/archive.go
@@ -185,7 +185,7 @@ func archiveFromDisk(directory string, src, dst string, allowDownload bool, excl
 		directory = filepath.Dir(directory)
 	}
 
-	options, err := archiveOptionsFor(directory, infos, dst, excludes, check)
+	options, err := archiveOptionsFor(directory, infos, dst, excludes, allowDownload, check)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


flouthoc (approved):
LGTM

PR #234, opened 2022-06-27
Use Travis for conformance tests

nalind:
Use Travis for conformance tests, instead of attempting to use GitHub Actions, which isn't enabled for this repository.

Run unit tests from "make test" using ./... as the package list; the dependency import tools are smart enough now to not pull in tests for our dependencies, so we don't need to try to grep out vendored packages.


flouthoc (approved):
LGTM
/lgtm

flouthoc:
@nalind ah i see. One thing which i saw is that CI was not green. It looks like conformance test takes lot of time and its breaching 30min timeout limit. Happens for me on the local as well. Should increase timeout for conformance test or maybe we can remove the specific tests which block conformance tests for such a long duration.

PR #236, opened 2022-09-05
add,copy: fix error message expecting at least two arguments

flouthoc:
Instruction `ADD` and `COPY` expects atlest two arguments instead of one
so change error message so imagebuilder spits correct error message.

Closes: https://github.com/containers/podman/issues/15628

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -136,7 +136,7 @@ func label(b *Builder, args []string, attributes map[string]bool, flagArgs []str
 //
 func add(b *Builder, args []string, attributes map[string]bool, flagArgs []string, original string) error {
 	if len(args) < 2 {
-		return errAtLeastOneArgument("ADD")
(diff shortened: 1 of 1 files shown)


dispatchers.go
```diff
@@ -704,6 +704,10 @@ func errAtLeastOneArgument(command string) error {
 	return fmt.Errorf("%s requires at least one argument", command)
 }
 
+func errAtLeastTwoArgument(command string) error {
+	return fmt.Errorf("%s requires at least two argument", command)
```
DaveJarvis, line 708: Typo: `arguments`

flouthoc: thanks fixed.

flouthoc:
Travis is just timing out, looks unrelated to this PR

rhatdan:
LGTM

TomSweeneyRedHat:
Late to the party, but LGTM

PR #237, opened 2022-09-12
Add support for RUN --network=none --security=insecure

rhatdan:
Docker now supports --network and --security option on the RUN line in Dockerfiles, this PR adds support for both.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -39,6 +39,10 @@ type Run struct {
 	Args  []string
 	// Mounts are mounts specified through the --mount flag inside the Containerfile
 	Mounts []string
+	// Network specifies the network mode to run the container with
+	Network string
+	// Security specifies the security mode to run the container with
+	Security string
 }
 
 type Executor interface {
@@ -79,7 +83,7 @@ func (logExecutor) Copy(excludes []string, copies ...Copy) error {
 }
 
 func (logExecutor) Run(run Run, config docker.Config) error {
-	log.Printf("RUN %v %v %t (%v)", run.Args, run.Mounts, run.Shell, config.Env)
+	log.Printf("RUN %v %v %v %v %t (%v)", run.Args, run.Mounts, run.Network, run.Security, run.Shell, config.Env)
 	return nil
 }
 
@@ -422,6 +426,10 @@ func (b *Builder) Run(step *Step, exec Executor, noRunsRemaining bool) error {
 	for _, run := range runs {
 		config := b.Config()
 		config.Env = step.Env
+		config.NetworkDisabled = (run.Network == "none")
+		if run.Security != "" {
+			config.SecurityOpts = append(config.SecurityOpts, run.Security)
+		}
 		if err := exec.Run(run, *config); err != nil {
 			return err
 		}
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -337,13 +337,14 @@ func run(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 
 	args = handleJSONArgs(args, attributes)
 
-	var mounts []string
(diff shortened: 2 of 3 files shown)


rhatdan:
@flouthoc @nalind PTAL

nalind:
The `dockerclient` implementation needs to complain about features it doesn't support.

rhatdan:
@nalind not sure where that happens?

nalind:
`dockerclient.ClientExecutor.Run()`.  Places where we've neglected this before produced #187, which was a bad experience for users who run the `imagebuilder` binary.

builder.go
```diff
@@ -422,6 +426,10 @@ func (b *Builder) Run(step *Step, exec Executor, noRunsRemaining bool) error {
 	for _, run := range runs {
 		config := b.Config()
 		config.Env = step.Env
+		config.NetworkDisabled = (run.Network == "none")
```
flouthoc, line 429: Shouldn't `config.NetworkDisabled` must be set on implementation end i.e buildah and imagebuilder should only parse `--network` and just set value in `run.Network` ? I always think of imagebuilder as the the frontend and buildah/docker as the actual implementation backend.

@nalind Could you correct if i got this wrong.

rhatdan: The only field to pass information to buildah is the NetworkDisabled field.  Even in the --security=insecure, there is no Privileged field to set, so I use the SecurityOpts to pass it through.

nalind: `imagebuilder` provides the parser in `dockerfile`, a builder in its top level package, and an executor that talks to an engine API (by way of go-dockerclient) that the builder can use in its `dockerclient` package.  Buildah uses the first two and supplies its own executor.

rhatdan:
@flouthoc Please take this over.  I am not going to have time to complete it

nalind:
I would also suggest dropping `--security`.  It's a "labs" feature in BuildKit, and not enabled there by default (compared to the current version of this PR, which makes it available all the time), and we don't have an identified use case for it.  What's the benefit to `imagebuilder`?

rhatdan:
The only use case would be if you wanted to run a --privileged container from within a build.  But I am fine with dropping it.

flouthoc:
I'm taking over this PR and will create two different PR for `RUN --network` and `RUN --security`

nalind:
I would again suggest dropping `--security`.

flouthoc:
> I would again suggest dropping `--security`.

@nalind Ackd dropping this for now.

nalind:
What are the use case and argument for overriding the specified or default value of the `--network` CLI setting from inside of a Dockerfile?  BuildKit at least requires that the user decide to allow these things at build-time, and I don't see any such precaution in here.

flouthoc:
> What are the use case and argument for overriding the specified or default value of the `--network` CLI setting from inside of a Dockerfile? BuildKit at least requires that the user decide to allow these things at build-time, and I don't see any such precaution in here.

I haven't check the PR implementation yet but I'd assume that `RUN --network` should only override for a specific `RUN` invocation and not for the entire build. So for instance a build can be invoked with CLI `--network=host` but certain `RUN` instructions can be isolated with `RUN --network=none <some task>`

rhatdan:
Yes that is what I would expect.

rhatdan:
When I originally looked at this, it looked like imagebuilder and buildah were setup to allow the override on a per RUN bases.

I guess if you were overly cautious you might have

RUN dnf -y update ...
RUN --network=none /run/local/script

flouthoc:
Should we close this one in favor of: https://github.com/openshift/imagebuilder/pull/247 and https://github.com/containers/buildah/pull/4566

rhatdan:
Sure.

PR #239, opened 2023-01-12
Use containers/storage/pkg/regexp to better handle init regex

rhatdan:
Use github.com/containers/storage/pkg/regexp to better handle init regex
    
    The new Reexp structure, will handle the initialization of regexp
    and then will do the MustCompile within a sync.Once field.
    
### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -12,7 +12,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


nalind:
LGTM, but unless there's a need to do otherwise, I prefer vendoring tagged versions of dependencies.  If we can wait for a tagged release of containers/storage that includes this API, I'd like to do that.

rhatdan:
I cut a release to move this along.

@flouthoc PTAL

flouthoc (approved):
LGTM

nalind:
Looks like #234 started triggering errors in Travis for a conformance test that was added in #222 but which was failing.  We may need to revisit #219.

rhatdan:
@giuseppe PTAL, looks like your change might be breaking tests?

rhatdan:
Looks like https://app.travis-ci.com/github/openshift/imagebuilder/jobs/593695010 is having the same problem.

nalind:
If I'm reading https://app.travis-ci.com/github/openshift/imagebuilder/jobs/595513477#L668 right, `docker build` produced an image which included `/var/www/file` and `/var/www/file2`, but the image that `imagebuilder` built using the same sources didn't include those.  But when I run the test with Go 1.20 and its version of moby-engine on my Fedora 37 box, the test passes.  Restarted the job to see if it's not just a weird flake.

PR #242, opened 2023-02-01
dispatcher,env: must honor heading `ARG` before `FROM`

flouthoc:
Correctly process any `ARG` if declared before assigning value to any `ENV`. This is important for use cases where `ARG` is configured and passed from the user.

Use case.

```Dockerfile
ARG CONT_IMG_VER
FROM ubuntu
ENV CONT_IMG_VER=${CONT_IMG_VER:-v1.0.0}
RUN echo $CONT_IMG_VER
```

```console
buildah build --build-arg CONT_IMG_VER=xyz -t test .
```

Closes: https://github.com/containers/buildah/issues/4547

### DIFF
--- a/builder_test.go
+++ b/builder_test.go
@@ -589,7 +589,7 @@ func TestBuilder(t *testing.T) {
 			Dockerfile: "dockerclient/testdata/Dockerfile.env",
 			From:       "busybox",
 			Config: docker.Config{
-				Env:   []string{"name=value", "name2=value2a            value2b", "name1=value1", "name3=value3a\\n\"value3b\"", "name4=value4a\\nvalue4b"},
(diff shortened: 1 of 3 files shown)


flouthoc:
@nalind @rhatdan @TomSweeneyRedHat PTAL. @nalind I think travis timeout is unrelated to this PR.

nalind:
https://app.travis-ci.com/github/openshift/imagebuilder/jobs/594889544#L654 suggests that the updated test case doesn't produce the same result between `imagebuilder` and `docker build`.

nalind:
Also, it might be worth seeing if bumping the timeout passed to `travis_wait` on line 19 of `.travis.yml` will avoid the timeout - the "external" conformance cases are some of the most time-consuming.

flouthoc:
Ah i think buildkit and docker does not support this ( I need to confirm this so holding till that )

PR #244, opened 2023-02-03
TestBuilder(): be able to expect errors

nalind:
Update some test cases so that we can fail a test if we expect to get an error, but don't see one.

### DIFF
--- a/builder_test.go
+++ b/builder_test.go
@@ -561,7 +561,8 @@ func TestBuilder(t *testing.T) {
 		Unrecognized []Step
 		Config       docker.Config
 		Image        *docker.Image
-		ErrFn        func(err error) bool
+		FromErrFn    func(err error) bool
+		RunErrFn     func(err error) bool
 	}{
 		{
 			Dockerfile: "dockerclient/testdata/dir/Dockerfile",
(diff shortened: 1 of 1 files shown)


builder_test.go
```diff
@@ -775,9 +776,6 @@ func TestBuilder(t *testing.T) {
 				Image: "centos:7",
 				Shell: []string{"/bin/bash", "-xc"},
 			},
-			ErrFn: func(err error) bool {
-				return err != nil && strings.Contains(err.Error(), "multiple FROM statements are not supported")
-			},
```
TomSweeneyRedHat, line 780: I'm not sure why this is getting removed?  More tea for me?

nalind: The Dockerfile the test uses doesn't have multiple FROM statements in it, so the function was never called.  Under the new paradigm, having a function that decides whether or not to ignore an error, but not having an error to pass to it, would cause the test to fail.

TomSweeneyRedHat: Thanks, thought this test had multiple FROM calls.  SGTM.

TomSweeneyRedHat:
LGTM

PR #241, opened 2023-01-20
Make conformance tests work on travis

saschagrunert:
This makes travis at least succeed on `master` again.

PR #245, opened 2023-02-06
Bump to Go 1.19

nalind:
Use Go 1.19 instead of 1.16 in go.mod.
Test with 1.19/1.20 in Travis instead of 1.17/1.18. Bump the Travis environment from the default (16.04) to jammy (22.04), which is the latest currently available in Travis.
Update assorted direct dependencies.

Obsoletes #241.

Update conformance tests nonroot-USER-before-WORKDIR-notused and nonroot-USER-before-WORKDIR-used to request that they be built with BuildKit.  Many conformance tests fail if we make this the default; we'll have to make that change and fix those later.


nalind:
> I will be surprised if this passes without [openshift/release#36040](https://github.com/openshift/release/pull/36040).

Well, I'm surprised.

rhatdan:
LGTM

nalind:
Hmm, updating to docker 23.0.0 apparently doesn't result in API calls getting BuildKit by default after all.  It must be a client default that changed.

rhatdan:
@flouthoc WDYT?

flouthoc:
> Hmm, updating to docker 23.0.0 apparently doesn't result in API calls getting BuildKit by default after all. It must be a client default that changed.

Is this a recent change in docker 23.0.0 ? the PR LGTM. @nalind I think this already includes the needed change which I added here https://github.com/openshift/imagebuilder/pull/246 hence closing #246

nalind:
> > Hmm, updating to docker 23.0.0 apparently doesn't result in API calls getting BuildKit by default after all. It must be a client default that changed.
> 
> Is this a recent change in docker 23.0.0 ?

It's the first item called out in the "New" section in https://docs.docker.com/engine/release-notes/23.0/.

rhatdan:
Since the tests pass is this ready to merge?

nalind:
> Since the tests pass is this ready to merge?

It was testing using a private fork of go-dockerclient to let the client request that the pair of cases that were failing be built using BuildKit instead, and I don't want us depending on my fork.  Now that https://github.com/fsouza/go-dockerclient/pull/961 is merged, we can avoid having to do that.

/retitle Bump to Go 1.19

nalind:
Also if anyone wants to approve https://github.com/openshift/release/pull/36040, it's sure to save us some trouble later.

rhatdan:
LGTM

nalind:
A reminder that this repository has a bot to handle merges.

PR #246, opened 2023-02-07
system: use `containers/storage` for `LCOW` validation

flouthoc:
Newer versions of `github.com/docker/docker` does not supports the LCOW validation functions which imagebuilder uses hence move them to a local package of imagebuilder and drop dependency on
`github.com/docker/docker/pkg/system`

Alternative solution: refactor imagebuilder to stop using `LCOWSupported()`

Closes and helps in: https://github.com/containers/buildah/pull/4555

### DIFF
--- a/dockerfile/parser/parser.go
+++ b/dockerfile/parser/parser.go
@@ -12,8 +12,8 @@ import (
 	"strings"
 	"unicode"
 
-	"github.com/docker/docker/pkg/system"
 	"github.com/openshift/imagebuilder/dockerfile/command"
+	"github.com/containers/storage/pkg/system"
 	"github.com/pkg/errors"
 )
 


flouthoc:
@nalind @rhatdan I think `moby` has dropped `LCOW` check and does not contain any public API for it, hence we can move it to imagebuilder or refactor imagebuilder so we can also drop `LCOW` support but I am not sure if anybody is still using logic for the former one hence a safer bet is to move `LCOW` checks here.

nalind:
That logic is still around in the containers/storage pkg/system.  Switch it to using that, it's less code to carry around in here.

flouthoc:
I see some overlapping work here: https://github.com/openshift/imagebuilder/pull/245 @nalind let me know if i can close this.

PR #247, opened 2023-02-08
builder,dispatcher: support inline `--network` in RUN statements

flouthoc:
Imagebuilder should allow clients to support inline --network in RUN stmts so users can create isolate or expose a particular build containers. 

```Dockerfile
FROM alpine
RUN --network=host wget google.com
RUN --network=none wget google.com
```

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -330,6 +330,7 @@ func run(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 	args = handleJSONArgs(args, attributes)
 
 	var mounts []string
+	var network string
 	filteredUserArgs := make(map[string]string)
 	for k, v := range b.Args {
 		if _, ok := b.AllowedArgs[k]; ok {
@@ -346,14 +347,17 @@ func run(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 		case strings.HasPrefix(arg, "--mount="):
 			mount := strings.TrimPrefix(arg, "--mount=")
 			mounts = append(mounts, mount)
+		case strings.HasPrefix(arg, "--network="):
+			network = strings.TrimPrefix(arg, "--network=")
 		default:
-			return fmt.Errorf("RUN only supports the --mount flag")
+			return fmt.Errorf("RUN only supports the --mount and --network flag")
 		}
 	}
 
 	run := Run{
-		Args:   args,
-		Mounts: mounts,
+		Args:    args,
+		Mounts:  mounts,
+		Network: network,
 	}
 
 	if !attributes["json"] {
--- a/dispatchers_test.go
+++ b/dispatchers_test.go
@@ -667,6 +667,35 @@ func TestDispatchRunFlags(t *testing.T) {
 
 }
 
+func TestDispatchNetworkFlags(t *testing.T) {
+	mybuilder := Builder{
+		RunConfig: docker.Config{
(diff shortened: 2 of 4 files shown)


rhatdan:
LGTM
@nalind PTAL

dispatchers_test.go
```diff
@@ -667,6 +667,36 @@ func TestDispatchRunFlags(t *testing.T) {
...
+			Shell:  true,
+			Args:   args,
+			Network: "none",
+		},
+	}
+
+	if !reflect.DeepEqual(mybuilder.PendingRuns, expectedPendingRuns) {
+		t.Errorf("Expected %v, to match %v\n", expectedPendingRuns, mybuilder.PendingRuns)
+	}
+
```
nalind, line 697: Extra blank line?

dispatchers.go
```diff
@@ -346,15 +347,23 @@ func run(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 		case strings.HasPrefix(arg, "--mount="):
 			mount := strings.TrimPrefix(arg, "--mount=")
 			mounts = append(mounts, mount)
+		case strings.HasPrefix(arg, "--network="):
+			network = strings.TrimPrefix(arg, "--network=")
 		default:
-			return fmt.Errorf("RUN only supports the --mount flag")
+			return fmt.Errorf("RUN only supports the --mount and --network flag")
+
```
nalind, line 354: Extra blank line?

nalind:
What does the `dockerclient` implementation do with this flag?  Quietly ignoring it would not be okay.

TomSweeneyRedHat:
LGTM once @nalind is hip with this.

flouthoc:
@nalind My intention is to wire it in the `Prepare(` step of `dockerclient` to simulate buildkit but now I think it will be an effort to simulate buildkit while imagebuilder does not actually invokes buildkit. So I am okay to remove or comment this line in case we want to implement this in future.

Ref: https://pkg.go.dev/github.com/fsouza/go-dockerclient#CreateContainerOptions

nalind:
`CreateExec()` and `StartExec()`, which the `dockerclient` implementation uses for handling RUN instructions, don't appear to me provide an option for altering the networking setup for the command being run.  I don't understand which line you're referring to removing or commenting out.  Do you mean the `Builder.Network` field that this currently adds?

flouthoc:
@nalind I have dropped that part, since it does not looks right, I will wire it when `go-dockerclient` actually supports buildkit.

nalind:
It's not clear what purpose the new `Builder.Network` field serves.
What does the `dockerclient` implementation do with this flag?  Quietly ignoring it would not be okay.

flouthoc:
@nalind My bad i removed the using line but forgot to remove the field, removed it now.

nalind:
Okay, good.  What does the `dockerclient` implementation do with this flag?  Quietly ignoring it would not be okay.

flouthoc:
@nalind I have added a failure for `RUN --network` just like its added for `RUN --mount` here: https://github.com/openshift/imagebuilder/commit/9c60e8980c52d9743f7298094f31f6527fae1e6e

flouthoc:
@nalind @rhatdan @TomSweeneyRedHat PTAL we can have `RUN --network` support in buildah once this gets merged.
/hold cancel

PR #248, opened 2023-02-09
dockerclient: don't quietly ignore --mount

nalind:
The dockerclient implementation doesn't support --mount.  Instead of quietly ignoring such a flag and hoping for the best (creating issues like #187), error out.


TomSweeneyRedHat:
LGTM

PR #249, opened 2023-02-09
dockerclient: implement COPY --chmod

### LINKED ISSUE #228: dockerclient quietly ignores the --chmod flag for ADD/COPY
We extended the parser to recognize it and add a field to `Copy` structs, but didn't teach `dockerclient` to either use the value or produce an error when people attempted to use the flag.

nalind:
We've been recognizing the argument for some time, we just ignored it.  Fixes #228.


dockerclient/client.go
```diff
@@ -1093,6 +1106,13 @@ func (e *ClientExecutor) CopyContainer(container *docker.Container, excludes []s
 				}
 				r = filtered
 			}
+			if c.Chmod != "" {
```
rhatdan, line 1109: Should this only happen if c.Chown != "" {?

nalind: Oof, no, it should not.

nalind: Wait, it doesn't...?

rhatdan: Your right the way github is displaying it confused me.

LGTM

TomSweeneyRedHat:
LGTM

PR #250, opened 2023-02-22
builder,volume: `Add` must return true if path pre-exists in set

flouthoc:
`func (s *VolumeSet) Add(path string) bool` must return `true` if its trying to add a path which already exists in the volume set.

When builder actually process volume it can add path to volume set but will not Preserve the path till a run is found.

Makes cases like this functional
```Dockerfile
FROM scratch
WORKDIR /bin

COPY hello /bin

ENTRYPOINT [ "hello" ]

VOLUME "/dev/urandom:/dev/urandom:--:iodev"

LABEL com.abc.product.rtp.rtpStackSize 0x1000000
LABEL com.abc.product.rtp.rtpPriority 50
LABEL com.abc.product.rtp.rtpOptions 0x80
LABEL com.abc.product.rtp.rtpTasksOptions 0x00
```

Where the output must contain volume `/dev`

Closes: https://github.com/containers/buildah/issues/4594

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -128,7 +128,13 @@ func (s *VolumeSet) Add(path string) bool {
 	path = strings.TrimSuffix(path, "/")
 	var adjusted []string
 	for _, p := range *s {
-		if p == path || strings.HasPrefix(path, p+"/") {
+		if p == path {
+			// if path is already in the set
+			// pretend it's added and return
+			// true
+			return true
+		}
+		if strings.HasPrefix(path, p+"/") {
 			return false
 		}
 		if strings.HasPrefix(p, path+"/") {
--- a/builder_test.go
+++ b/builder_test.go
@@ -28,8 +28,8 @@ func TestVolumeSet(t *testing.T) {
 		uncovered []string
 	}{
 		{
-			inputs:  []string{"/var/lib", "/var"},
-			changed: []bool{true, true},
+			inputs:  []string{"/var/lib", "/var", "/var"},
+			changed: []bool{true, true, true},
 			result:  []string{"/var"},
 
 			covered:   []string{"/var/lib", "/var/", "/var"},
--- a/dockerclient/conformance_test.go
+++ b/dockerclient/conformance_test.go
@@ -396,6 +396,10 @@ func TestConformanceInternal(t *testing.T) {
 			Dockerfile: "testdata/Dockerfile.shell",
 		},
 		{
+			Name:       "preparepathwithoutrun",
+			Dockerfile: "testdata/Dockerfile.preparewithoutrun",
+		},
+		{
 			Name:       "args",
 			Dockerfile: "testdata/Dockerfile.args",
 			Args:       map[string]string{"BAR": "first"},
--- a/dockerclient/testdata/Dockerfile.preparewithoutrun
+++ b/dockerclient/testdata/Dockerfile.preparewithoutrun
@@ -1,0 +1,5 @@
+FROM busybox
+VOLUME "/hello"
+# This grep will fail if hello path does not persists
+RUN ls -a | grep hello
+LABEL ho hey


flouthoc:
@nalind @TomSweeneyRedHat @rhatdan PTAL

nalind:
The argument value for VOLUME is a filesystem path, so `VOLUME "/dev/urandom:/dev/urandom:--:iodev"` doesn't make sense.  That makes it hard to understand what you're trying to do here.

flouthoc:
@nalind Sorry for confusion I copied exact example from the issue, here buildah precreates a `dev` inside the image in first example it does not

** fails**
```Dockerfile
FROM scratch
WORKDIR /bin

COPY hello /bin

ENTRYPOINT [ "hello" ]

VOLUME "/dev"

LABEL com.abc.product.rtp.rtpStackSize 0x1000000
LABEL com.abc.product.rtp.rtpPriority 50
LABEL com.abc.product.rtp.rtpOptions 0x80
LABEL com.abc.product.rtp.rtpTasksOptions 0x00
```

** works**
```Dockerfile
FROM scratch
WORKDIR /bin

COPY hello /bin

ENTRYPOINT [ "hello" ]

LABEL com.abc.product.rtp.rtpStackSize 0x1000000
LABEL com.abc.product.rtp.rtpPriority 50
LABEL com.abc.product.rtp.rtpOptions 0x80
LABEL com.abc.product.rtp.rtpTasksOptions 0x00
VOLUME "/dev"
```

nalind:
Ah.  The behavior of when exactly a directory is created for a volume, and when it isn't, is something we'd _need_ to have conformance tests for.

nalind:
We've gotten burned pretty badly by waiting to add conformance tests until after the things they test were merged, so please add them along with the change they test.

builder.go
```diff
@@ -128,7 +128,13 @@ func (s *VolumeSet) Add(path string) bool {
 	path = strings.TrimSuffix(path, "/")
 	var adjusted []string
 	for _, p := range *s {
-		if p == path || strings.HasPrefix(path, p+"/") {
+		if p == path {
+			// if path is already in the set
+			// pretend its added and return
```
TomSweeneyRedHat, line 133: nit of a nit of a nit.  Only worry about if you have other changes.
```suggestion
			// pretend it's added and return
```

builder_test.go
```diff
@@ -28,8 +28,8 @@ func TestVolumeSet(t *testing.T) {
 		uncovered []string
 	}{
 		{
-			inputs:  []string{"/var/lib", "/var"},
-			changed: []bool{true, true},
+			inputs:  []string{"/var/lib", "/var", "/var"},
+			changed: []bool{true, true, true},
```
nalind, line 32: Why is deduplication not desired here?

nalind:
The commit log still includes a nonsensical VOLUME instruction in its description of the problem.

I'm still not seeing a connection between what the new conformance test does and the description of the problem this is solving - the Dockerfile in question has only one VOLUME instruction in it.  When I try what I think would match the description, this:
```Dockerfile
FROM busybox as base
VOLUME /var
FROM base
VOLUME /var
VOLUME /var/lib
```
both `docker build` and `imagebuilder` build an image without complaining.

rhatdan:
@flouthoc Reminder about this PR?

shengjiangwuqq:
Hi all,

I have tried this commit on my local ubuntu machine, it does **NOT** solve this issue 

https://github.com/containers/buildah/issues/4594

Any comment? Thanks.

shengjiangwuqq:
Hello, is anyone working on this?

flouthoc:
Hi @shengjiangwuqq Thanks for testing this out, I will check this but this is on a comparatively lower priority since I am caught up with some other things. I will get back to this somewhere in next month.

rhatdan:
@flouthoc Lets get this in.

flouthoc:
Something was not correct in this PR as suggested here https://github.com/openshift/imagebuilder/pull/250#issuecomment-1462121866 I'll revisit this.

PR #256, opened 2023-05-16
RUN-1833: dockerclient: use /bin/sleep for idling instead of sleep

nalind:
The built-in "sleep" in the latest alpine base image's shell behaves differently from the version it has to exec(), exiting when it receives a stray SIGCHLD for one of the processes we're running in an exec session, and this is causing a consistent failure in the "copy and env interaction" conformance test.
Call /bin/sleep explicitly to work around this.

Note to other maintainers: please do not /lgtm until Travis is green.


TomSweeneyRedHat:
LGTM
and happy green test buttons.

PR #261, opened 2023-07-24
dockerfile/parser: use custom 2mb buffer for long-texts 

flouthoc:
Containerfile may contain long lines lets use custom buffer for such
use-cases.

Closes: https://github.com/containers/buildah/issues/4929


flouthoc:

rhatdan:
LGTM
@nalind PTAL

dockerfile/parser/parser.go
```diff
@@ -312,6 +316,11 @@ func Parse(rwc io.Reader) (*Result, error) {
 		root.AddChild(child, startLine, currentLine)
 	}
 
+	scannerErr := scanner.Err()
+	if scannerErr != nil {
```
nalind, line 320: Go ahead and use the `if scannerErr := scanner.Err(); scannerErr != nil {` form here, since `scannerErr` isn't going to be used anywhere else.

nalind:
One nit, otherwise LGTM.

PR #262, opened 2023-08-01
builder: support variables in stage names

SoMuchForSubtlety:
Sometimes it can be useful to have dynamic stage names.

I'm not sure if my change to `dispatchers.go` is correct. Previously, the `flagArgs` got processed without `HeadingArgs` and I wasn't sure if this omission was on purpose or a bug. Please let me know if you want me to undo this change.
 
Closes https://github.com/containers/buildah/issues/4948

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -235,7 +235,10 @@ func NewStages(node *parser.Node, b *Builder) (Stages, error) {
 		return stages, err
 	}
 	for i, root := range SplitBy(node, command.From) {
-		name, _ := extractNameFromNode(root.Children[0])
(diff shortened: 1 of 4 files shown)


flouthoc:
Discussing here: https://github.com/containers/buildah/issues/4948#issuecomment-1670952849 since this is not in parity with `docker/buildkit` so this feature will be applicable only for `Containerfile` and not `Dockerfiles`

rhatdan:
@nalind PTAL

rhatdan:
Since we have had no activity on this for many months closing.

PR #264, opened 2023-10-19
Containerfile: Add `heredoc` support for `RUN`, `COPY` and `ADD`

flouthoc:
Imagebuilder must honor and support parsing heredoc syntax from `RUN, COPY and ADD` and allow implementers to implement it.

Will help in: https://github.com/containers/buildah/issues/3474


dispatchers.go
```diff
@@ -313,6 +315,17 @@ func workdir(b *Builder, args []string, attributes map[string]bool, flagArgs []s
 	return nil
 }
 
+func parseCommandForHeredoc(firstline string) string {
+	// Expected format `RUN <str> <<EOF`
+	if strings.HasPrefix(firstline, "RUN ") {
+		firstline = strings.TrimPrefix(firstline, "RUN ")
+		if strings.HasSuffix(firstline, "<<EOF") {
```
rhatdan, line 322: Should we support <<eof as well?  Just add a ToUpper to Firstline?

rhatdan: Or should we generalize this totally and allow `<< foobar`

rhatdan: Why is this only RUN, shouldn't this handle COPY and ADD?
Since RUN can be lowercase, this should also use ToUpper.

flouthoc: @rhatdan Yes PR is still WIP, I need to extended this to `COPY` and `ADD` and yes this should be not restricted to `<<EOF` it should work with any identifier. The code is pretty much in draft and still under progress.

flouthoc: Extended for `ADD` and `COPY` as well.

flouthoc:
Following PR is ready for review, please refer to `dispatchers_test.go` and `dockerfile/parser/parser_test.go` for newly added tests and some trivial tests are not added as some of the code is imported directly from buildkit and they are tested there.

@rhatdan @nalind PTAL

flouthoc:
Failing conformance is just timeout, other conformance test for `Go 1.20` passed.

dispatchers.go
```diff
@@ -354,10 +404,26 @@ func run(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 		}
 	}
 
+	var files []File
+	for _, heredoc := range heredocs {
+		command := parseCommandForHeredoc(original)
```
rhatdan, line 409: Should we check for if command==""

flouthoc: No its not needed, empty command is expected and code further honors it, there are test in `dispatcher_test.go` for both the cases and further test in buildah https://github.com/containers/buildah/pull/5092


Here empty command only means that treat default shell or `sh` as the default command.


For instance in command is parsed as empty which means implementation should use default shell and command.
```Dockerfile
RUN <<EOF
echo "Hello" >> /hello
echo "World!" >> /hello
EOF
```

Here command is parsed as `python3` instead of default shell
```Dockerfile
RUN python3 <<EOF
with open("/hello", "w") as f:
    print("Hello", file=f)
    print("Something", file=f)
EOF
```

builder.go
```diff
@@ -31,6 +32,14 @@ type Copy struct {
...
 	Chown string
 	Chmod string
+	IsHeredoc bool
+	Files   []File
+}
+
+type File struct {
+	Command string
+	Name    string
+	Data    string
```
nalind, line 42: These need some godoc to indicate to a caller what they're supposed to do with these new fields.

builder.go
```diff
@@ -31,6 +32,14 @@ type Copy struct {
...
 }
```
nalind, line 43: These need some godoc to indicate to an `Executor` implementation what it's supposed to do with these new fields.

nalind:
Is the `imagebuilder` CLI going to start quietly ignoring this syntax, or is it going to continue to flag it as an error because it doesn't understand it?

dispatchers.go
```diff
@@ -164,14 +166,32 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
...
+			Data:    content,
+			Name:    heredoc.Name,
+			Command: command,
+		}
+		files = append(files, file)
+	}
+	isheredoc := false
+	if len(files) > 0 {
+		isheredoc = true
+	}
```
nalind, line 186: This chunk is copied verbatim in three places.  It should be a helper function that also takes the `command` as a parameter, since each caller already knows exactly what its value should be.

nalind: Oh, `command` is not the instruction.

dispatchers.go
```diff
@@ -313,6 +351,18 @@ func workdir(b *Builder, args []string, attributes map[string]bool, flagArgs []s
 	return nil
 }
 
+func parseCommandForHeredoc(firstline string) string {
+	// Expected format `RUN <str> <<`
+	if strings.HasPrefix(firstline, "RUN ") {
+		firstline = strings.TrimPrefix(firstline, "RUN ")
```
nalind, line 357: This is case sensitive, while other parts of the parser are not.

dockerfile/parser/parser.go
```diff
@@ -85,6 +110,7 @@ var (
 	tokenWhitespace      = sRegexp.Delayed(`[\t\v\f\r ]+`)
 	tokenEscapeCommand   = sRegexp.Delayed(`^#[ \t]*escape[ \t]*=[ \t]*(?P<escapechar>.).*$`)
 	tokenPlatformCommand = sRegexp.Delayed(`^#[ \t]*platform[ \t]*=[ \t]*(?P<platform>.*)$`)
+	reHeredoc            = regexp.MustCompile(`^(\d*)<<(-?)([^<]*)$`)
```
nalind, line 113: Any reason this is using `regexp.MustCompile()` rather than `sRegexp.Delayed()` like the rest?

dockerfile/parser/parser.go
```diff
@@ -331,6 +404,26 @@ func Parse(rwc io.Reader) (*Result, error) {
 	}, nil
 }
 
+func heredocsFromLine(line string) ([]buildkitparser.Heredoc, error) {
```
nalind, line 407: Did this previously use `reHeredoc` instead of calling out to new dependencies?  I can't see the regexp being used anywhere.

dockerfile/parser/testfiles/heredoc/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+EOF
+
+COPY <<robots.txt <<humans.txt /test/
+(robots content)
+robots.txt
+(humans content)
+humans.txt
+
+RUN cat index.html
+RUN cat hello
```
nalind, line 25: I'd like to see some test coverage of the `FD<<word`, `<<-`, and `<<"word"` variations, both by themselves and in combinations.

flouthoc: Done, for othercases but i think `<<"word"` is not a valid case ( it does not works with buildkit as well I have verified manually ). For all othercases extended the test.

dockerfile/parser/parser.go
flouthoc: This variable is not needed. I have removed it.

dockerfile/parser/parser.go
flouthoc: Yes its not needed anymore.

dispatchers.go
flouthoc: Fixed.

dispatchers.go
flouthoc: Fixed.

builder.go
flouthoc: Done.

flouthoc:
> Is the `imagebuilder` CLI going to start quietly ignoring this syntax, or is it going to continue to flag it as an error because it doesn't understand it?

Did you mean `dockerclient` , I have added error for dockerclient implementations.

dispatchers.go
```diff
@@ -313,6 +349,22 @@ func workdir(b *Builder, args []string, attributes map[string]bool, flagArgs []s
...
+		} else if strings.HasPrefix(firstline, "run ") {
+			firstline = strings.TrimPrefix(firstline, "run ")
+		}
+		if strings.Contains(firstline, "<<") {
+			arr := strings.Split(firstline, "<<")
+			return strings.TrimSpace(arr[0])
+		}
+	}
+	return ""
+}
```
TomSweeneyRedHat, line 366: Are you guaranteed not to have a space in front of "run" or "RUN" at this point? I think so, but things can go awry here if not.

flouthoc: Yes at this point this RUN instruction is already parsed, anything other than this will be rejected by the parser in earlier cycles.

nalind (changes requested):

dispatchers.go
flouthoc: Added test cases for these in `dispatcher_test.go` in function `TestParseHeredocCommand`

dockerfile/parser/testfiles/heredoc/Dockerfile
flouthoc: My bad I misunderstood with terminator also having quotes i.e `"EOF"`. The dockerfile which you have pasted is valid and I have added a case for the Dockerfile which you have provided.

The bad dockerfile which i was trying
```Dockerfile
# syntax=docker/dockerfile:1.4
FROM alpine
RUN <<"EOF"
id > /tmp/id.txt
"EOF"
```

builder.go
```diff
@@ -29,8 +30,21 @@ type Copy struct {
 	Download bool
 	// If set, the owner:group for the destination.  This value is passed
 	// to the executor for handling.
-	Chown string
-	Chmod string
+	Chown     string
+	Chmod     string
+	IsHeredoc bool
```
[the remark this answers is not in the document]
flouthoc: My idea was to not keep `Files` restricted to only heredocs and let it be open for other use-cases which might show up in future but that's too much thinking i have removed it for now and only using `len(Files)` for such checks.

builder.go
```diff
@@ -29,8 +30,21 @@ type Copy struct {
...
+	Files []File
+}
+
+// File defines if any additional file needs to be created
+// by the executor instruction so that specified command
+// can execute/copy the created file inside the build container.
+type File struct {
+	Command string // Executor can read this field to override the default command.
+	Name    string // Name of the new file.
+	Data    string // Content of the file.
```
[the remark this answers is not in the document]
flouthoc: Added Filedescriptor as well, but afaics in buildkit FileDescriptor is defined in struct but it is never used. Anyways we should still process it so added it to struct.

dispatchers.go
```diff
@@ -201,14 +222,29 @@ func dispatchCopy(b *Builder, args []string, attributes map[string]bool, flagArg
...
+	for _, heredoc := range heredocs {
+		command := parseCommandForHeredoc(original)
+		content := heredoc.Content
+		if heredoc.Chomp {
+			content = buildkitparser.ChompHeredocContent(content)
+		}
+		file := File{
+			Data:    content,
+			Name:    heredoc.Name,
+			Command: command,
```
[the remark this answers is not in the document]
flouthoc: Afaics `heredoc.FileDescriptor` is defined in buildkit struct but even buildkit never uses it ( correct me if i am looking code wrongly ) but I have added it to imagebuilder and buildah can implement it even if buildkit does not implements it.

I am using `heredoc.Expand` in imagebuilder itself for expansion of heredoc instead of letting buildah do it, in function `processHereDocs` in `dispatcher.go`

flouthoc:
@nalind Could you PTAL again.

TomSweeneyRedHat:
@flouthoc tests aren't hip

flouthoc:
> Failing conformance is just timeout, other conformance test for `Go 1.20` passed.

@TomSweeneyRedHat Is just a timeout which is unrelated to this PR.

builder.go
```diff
@@ -31,6 +32,21 @@ type Copy struct {
...
+	// instruction.
+	Files []File
+}
+
+// File defines if any additional file needs to be created
+// by the executor instruction so that specified command
+// can execute/copy the created file inside the build container.
+type File struct {
+	FileDescriptor uint   // If output needs to be redirected to any FileDescriptor
+	Expand         bool   // Needs to expand heredoc `Data` with build args or environment variable
```
nalind, line 45: We don't set this flag, and seem to only provide already-expanded values of `Data`, so I think we can drop this field.

dispatchers.go
```diff
@@ -164,14 +195,18 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 			return fmt.Errorf("ADD only supports the --chmod=<permissions> and the --chown=<uid:gid> flag")
 		}
 	}
-	b.PendingCopies = append(b.PendingCopies, Copy{Src: args[0:last], Dest: dest, Download: true, Chown: chown, Chmod: chmod})
+	files, err := processHereDocs(original, heredocs, args)
```
nalind, line 198: Should this be passing `userArgs` instead of `args`?

dispatchers.go
```diff
@@ -201,14 +236,18 @@ func dispatchCopy(b *Builder, args []string, attributes map[string]bool, flagArg
 			return fmt.Errorf("COPY only supports the --chmod=<permissions> --chown=<uid:gid> and the --from=<image|stage> flags")
 		}
 	}
-	b.PendingCopies = append(b.PendingCopies, Copy{From: from, Src: args[0:last], Dest: dest, Download: false, Chown: chown, Chmod: chmod})
+	files, err := processHereDocs(original, heredocs, args)
```
nalind, line 239: Should this be passing `userArgs` instead of `args`?

dispatchers.go
```diff
@@ -354,10 +445,16 @@ func run(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 		}
 	}
 
+	files, err := processHereDocs(original, heredocs, args)
```
nalind, line 448: Should this be passing `userArgs` instead of `args`?

dispatchers.go
nalind: Are we testing syntax like this example from that section?
```Dockerfile
RUN <<FILE1 cat > file1 && <<FILE2 cat > file2
I am
first
FILE1
I am
second
FILE2
```

dockerfile/parser/testfiles/heredoc/Dockerfile
nalind: We still need at least one test where the descriptor number isn't 0, something like
```Dockerfile
RUN 5<<EOF cat /proc/self/fd/5 > file.txt
this is the file
EOF
```

nalind (changes requested):

dockerfile/parser/testfiles/heredoc/Dockerfile
flouthoc: Done.

dispatchers.go
flouthoc: Added test cases for this with changed code.

flouthoc:
@nalind PTAL

dispatchers.go
```diff
@@ -313,6 +352,90 @@ func workdir(b *Builder, args []string, attributes map[string]bool, flagArgs []s
...
+					lasttolastToken := commandSplit[len(commandSplit)-2]
+					command = lastToken
+					if lasttolastToken == "&&" || lasttolastToken == "||" {
+						command = lasttolastToken + " " + command
+					}
+				}
+			}
+		}
+		// check if any piped instruction
+		tokens := strings.Split(originalCommand, " ")
```
nalind, line 389: Does this handle quoted whitespace in the passed-in line?  Can we add a test for that?

nalind (changes requested):

dispatchers.go
flouthoc: Added a test case in `TestParseHeredocCommand`

dockerfile/parser/testfiles/heredoc/Dockerfile
flouthoc: Done added few more cases which should cover this.

flouthoc:
@nalind Could you PTAL again.

dispatchers_test.go
```diff
@@ -790,11 +916,110 @@ func TestDispatchFromFlagsAndUseBuiltInArgs(t *testing.T) {
...
 
 	if mybuilder.Platform != expectedPlatform {
 		t.Errorf("Expected %v, to match %v\n", expectedPlatform, mybuilder.Platform)
 	}
 }
+
+func TestParseCommandForHeredoc(t *testing.T) {
+	command, pipe, err := parseCommandForHeredoc("RUN python3 <<EOF", "EOF", 0)
+	require.NoError(t, err)
+	assert.Equal(t, command, "python3")
```
nalind, line 931: `assert.Equal()` expects to be passed the expected value before the actual value, and if it isn't, the error message gets very confusing.  This has tripped me up more than a few times.

nalind (changes requested):

dockerfile/parser/parser.go
```diff
@@ -331,6 +403,26 @@ func Parse(rwc io.Reader) (*Result, error) {
...
+	shlex := buildkitshell.NewLex('\\')
+	shlex.RawQuotes = true
+	shlex.RawEscapes = true
+	shlex.SkipUnsetEnv = true
+	words, _ := shlex.ProcessWords(line, []string{})
```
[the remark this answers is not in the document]
flouthoc: It has ability to preserve RawQuotes and Escapes as it is which i think is missing in imagebuilder's process word's. I think imagebuilder's ProcessWords can be extended to match these features but it will be easier in a new PR.

flouthoc:
Please don't review yet, this needs major refactor, I found some issues and code can be reduced here.

dispatchers_test.go
```diff
@@ -790,11 +916,79 @@ func TestDispatchFromFlagsAndUseBuiltInArgs(t *testing.T) {
...
+	if err := from(&mybuilder, args, nil, flags, original, nil); err != nil {
 		t.Errorf("dispatchAdd error: %v", err)
 	}
 
 	if mybuilder.Platform != expectedPlatform {
 		t.Errorf("Expected %v, to match %v\n", expectedPlatform, mybuilder.Platform)
 	}
 }
+
+func TestParseHeredocCommand(t *testing.T) {
```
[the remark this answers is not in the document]
flouthoc: This code is removed.

dispatchers_test.go
```diff
@@ -790,11 +916,79 @@ func TestDispatchFromFlagsAndUseBuiltInArgs(t *testing.T) {
...
+
+	command, pipe, err = parseCommandForHeredoc("RUN <<FILE1 cat > file1 && <<FILE2 cat > file2 && python3 <<FILE3 > file3", "FILE3")
+	require.NoError(t, err)
+	assert.Equal(t, command, "&& python3")
+	assert.Equal(t, pipe, "file3")
+
+	command, pipe, err = parseCommandForHeredoc("RUN <<FILE1 cat > file1 && echo hello world && echo something", "FILE1")
+	require.NoError(t, err)
+	assert.Equal(t, command, "cat")
+	assert.Equal(t, pipe, "file1 && echo hello world && echo something")
```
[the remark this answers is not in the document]
flouthoc: This code is removed.

dockerfile/parser/parser.go
```diff
@@ -53,7 +56,11 @@ func (node *Node) Dump() string {
 	for _, n := range node.Children {
 		str += "(" + n.Dump() + ")\n"
 	}
-
+	if len(node.Heredocs) > 0 {
+		for _, doc := range node.Heredocs {
+			str += "(" + doc.Name + "-" + doc.Content + ")\n"
```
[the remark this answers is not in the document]
flouthoc: This code is removed and not needed.

dispatchers.go
```diff
@@ -313,6 +352,90 @@ func workdir(b *Builder, args []string, attributes map[string]bool, flagArgs []s
...
+	// Expected format `RUN <str> <<`
+	firstline := originalCommand
+	command := ""
+	pipeCommand := ""
+	if strings.HasPrefix(firstline, "RUN ") || strings.HasPrefix(firstline, "run ") {
+		if strings.HasPrefix(firstline, "RUN ") {
+			firstline = strings.TrimPrefix(firstline, "RUN ")
+		} else if strings.HasPrefix(firstline, "run ") {
+			firstline = strings.TrimPrefix(firstline, "run ")
+		}
```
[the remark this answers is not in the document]
flouthoc: This code is removed and not needed.

dispatchers.go
flouthoc: This code is not needed.

dispatchers.go
```diff
@@ -313,6 +353,92 @@ func workdir(b *Builder, args []string, attributes map[string]bool, flagArgs []s
...
 
+func parseCommandForHeredoc(originalCommand string, name string, fd uint) (string, string, error) {
+	// Expected format `RUN <str> <<`
+	name = strings.ToLower(name)
+	originalCommand = strings.ToLower(originalCommand)
+	firstline := originalCommand
+	command := ""
+	pipeCommand := ""
+	if strings.HasPrefix(firstline, "run ") {
+		firstline = strings.TrimPrefix(firstline, "run ")
```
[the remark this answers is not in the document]
flouthoc: This code is removed and not needed.

dispatchers.go
```diff
@@ -313,6 +353,92 @@ func workdir(b *Builder, args []string, attributes map[string]bool, flagArgs []s
...
+		tokens := strings.Split(originalCommand, " ")
+		tokensAfterTerm := []string{}
+		foundHeredocTerminator := false
+		for _, token := range tokens {
+			if strings.HasPrefix(token, splitSep) {
+				foundHeredocTerminator = true
+				continue
+			}
+			if foundHeredocTerminator {
+				if strings.HasPrefix(token, "<<") || strings.HasPrefix(token, "<<-") || heredocTerm.Match([]byte(token)) {
```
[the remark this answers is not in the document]
flouthoc: This code is removed and not needed.

builder.go
```diff
@@ -31,6 +32,20 @@ type Copy struct {
...
+	// instruction.
+	Files []File
+}
+
+// File defines if any additional file needs to be created
+// by the executor instruction so that specified command
+// can execute/copy the created file inside the build container.
+type File struct {
+	FileDescriptor uint   // If output needs to be redirected to any FileDescriptor
+	PipeOutput     string // If output of heredoc needs to be piped to another command.
```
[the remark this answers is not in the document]
flouthoc: This code is removed and not needed.

dispatchers_test.go
flouthoc: This code is removed and not needed.

flouthoc:
Added integration test for all complex use-cases in the buildah PR as well: https://github.com/containers/buildah/pull/5092

flouthoc:
@nalind PTAL

dispatchers_test.go
```diff
@@ -90,6 +91,129 @@ func TestDispatchCopy(t *testing.T) {
...
+	heredocs := []buildkitparser.Heredoc{{Name: "EOF", Expand: true, Content: "(your index page goes here)"}}
+	if err := add(&mybuilder, args, nil, flagArgs, original, heredocs); err != nil {
+		t.Errorf("add error: %v", err)
+	}
+
+	expectedPendingCopies := []Copy{
+		{
+			From:     "",
+			Src:      []string{"<<EOF"},
+			Dest:     "/index.html", // destination must contain a trailing slash
```
nalind, line 146: This comment is a bit confusing, since the value for this field does not contain a trailing slash.

dispatchers_test.go
```diff
@@ -90,6 +91,129 @@ func TestDispatchCopy(t *testing.T) {
...
+		},
+	}
+
+	flags := []string{}
+	args := []string{"<< EOF"}
+	original := "RUN <<EOF"
+	heredocs := []buildkitparser.Heredoc{{Name: "EOF", Content: "echo \"Hello\" >> /hello\necho \"World\" >> /hello"}}
+
+	if err := run(&mybuilder, args, nil, flags, original, heredocs); err != nil {
+		t.Errorf("add error: %v", err)
```
nalind, line 171:
```suggestion
		t.Errorf("run error: %v", err)
```

dispatchers_test.go
```diff
@@ -680,13 +804,13 @@ func TestDispatchNetworkFlags(t *testing.T) {
 	args := []string{"echo \"stuff\""}
 	original := "RUN --network=none echo \"stuff\""
 
-	if err := run(&mybuilder, args, nil, flags, original); err != nil {
+	if err := run(&mybuilder, args, nil, flags, original, nil); err != nil {
 		t.Errorf("dispatchAdd error: %v", err)
```
nalind, line 808:
```suggestion
		t.Errorf("run error: %v", err)
```

dispatchers_test.go
```diff
@@ -715,7 +839,7 @@ func TestDispatchRunFlagsWithArgs(t *testing.T) {
 	args := []string{"echo \"stuff\""}
 	original := "RUN --mount=type=${TYPE},target=/foo echo \"stuff\""
 
-	if err := run(&mybuilder, args, nil, flags, original); err != nil {
+	if err := run(&mybuilder, args, nil, flags, original, nil); err != nil {
 		t.Errorf("dispatchAdd error: %v", err)
```
nalind, line 843:
```suggestion
		t.Errorf("run error: %v", err)
```

dispatchers_test.go
```diff
@@ -737,7 +861,7 @@ func TestDispatchRunFlagsWithArgs(t *testing.T) {
 			Image:      "busybox",
 		},
 	}
-	if err := run(&mybuilder, args, nil, flags, original); err != nil {
+	if err := run(&mybuilder, args, nil, flags, original, nil); err != nil {
 		t.Errorf("dispatchAdd error: %v", err)
```
nalind, line 865:
```suggestion
		t.Errorf("run error: %v", err)
```

dispatchers_test.go
```diff
@@ -767,7 +891,7 @@ func TestDispatchFromFlags(t *testing.T) {
 	args := []string{""}
 	original := "FROM --platform=linux/arm64 busybox"
 
-	if err := from(&mybuilder, args, nil, flags, original); err != nil {
+	if err := from(&mybuilder, args, nil, flags, original, nil); err != nil {
 		t.Errorf("dispatchAdd error: %v", err)
```
nalind, line 895:
```suggestion
		t.Errorf("from error: %v", err)
```

dispatchers_test.go
```diff
@@ -790,7 +914,7 @@ func TestDispatchFromFlagsAndUseBuiltInArgs(t *testing.T) {
 	args := []string{""}
 	original := "FROM --platform=$BUILDPLATFORM busybox"
 
-	if err := from(&mybuilder, args, nil, flags, original); err != nil {
+	if err := from(&mybuilder, args, nil, flags, original, nil); err != nil {
 		t.Errorf("dispatchAdd error: %v", err)
```
nalind, line 918:
```suggestion
		t.Errorf("from error: %v", err)
```

dockerfile/parser/parser.go
```diff
@@ -313,6 +352,39 @@ func Parse(rwc io.Reader) (*Result, error) {
...
+					}
+					if bytes.Equal(possibleTerminator, terminator) {
+						terminated = true
+						break
+					}
+					heredoc.Content += "\n"
+					heredoc.Content += string(bytesRead)
+				}
+				if !terminated {
+					return nil, errors.New("unterminated heredoc")
```
nalind, line 381: I'd suggest including `heredoc.Name` in the error message.

dockerfile/parser/testfiles/heredoc/Dockerfile
nalind: All of the tests in here include a blank line after the terminator, so we're not going to notice if we start skipping the next line after the terminator when it isn't blank.  Please add a case where the final heredoc terminator for an instruction line is immediately followed by a non-empty line.

dispatchers_test.go
flouthoc: removed comment it was copied from another block

dispatchers_test.go
flouthoc: done.

dockerfile/parser/parser.go
flouthoc: done

dockerfile/parser/testfiles/heredoc/Dockerfile
flouthoc: done.

flouthoc:
@nalind PTAL

dispatchers_test.go
```diff
@@ -679,7 +803,7 @@ func TestDispatchRunFlags(t *testing.T) {
 	args := []string{"echo \"stuff\""}
 	original := "RUN --mount=type=bind,target=/foo echo \"stuff\""
 
-	if err := run(&mybuilder, args, nil, flags, original); err != nil {
+	if err := run(&mybuilder, args, nil, flags, original, nil); err != nil {
 		t.Errorf("dispatchAdd error: %v", err)
```
nalind, line 807:
```suggestion
		t.Errorf("run error: %v", err)
```

nalind: Please do a pass checking that `Errorf` messages match the function that produced the assorted errors.  I'm only easily able to comment on lines that this PR touches.

nalind:
LGTM, with nits about error messages in tests that are mistaken about which function produced the errors that they're complaining about.

flouthoc:
@nalind @rhatdan Done, PTAL again

TomSweeneyRedHat:
Woot!  Very well done @flouthoc , great work here!

PR #265, opened 2023-11-06
builder,dispatcher: Parse ADD --checksum

jfroy:
Parse the new ADD --checksum flag.

See https://docs.docker.com/build/dockerfile/release-notes/#160.

This will be consumed by buildah. See containers/buildah#5135.

```Dockerfile
FROM scratch
ADD --checksum=sha256:5df80a9219a28d21f96e8c99e5406d19dbe65b9673a7724e587584b02143038c https://github.com/containers/podman/releases/download/v4.7.2/podman-remote-static-linux_amd64.tar.gz /
```

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -29,8 +29,9 @@ type Copy struct {
 	Download bool
 	// If set, the owner:group for the destination.  This value is passed
 	// to the executor for handling.
-	Chown string
-	Chmod string
+	Chown    string
+	Chmod    string
+	Checksum string
 }
 
 // Run defines a run operation required in the container.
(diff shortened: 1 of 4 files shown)


nalind:
Thanks for this!

/ok-to-test

Can you have the `dockerclient.ClientExecutor` implementation (somewhere around `dockerclient/client.go`'s line 879) complain when its methods are passed new fields which it doesn't know what to do with, sort of like the `Run()` method complains about flags that it doesn't implement?  If we don't remember to do that, it'll just start quietly ignoring flags that it used to produce errors over before, and users of the `imagebuilder` CLI might come to think that those options actually work.  That is an oversight that has come back to bite us before.

jfroy:
> Thanks for this!
> 
> /ok-to-test
> 
> Can you have the `dockerclient.ClientExecutor` implementation (somewhere around `dockerclient/client.go`'s line 879) complain when its methods are passed new fields which it doesn't know what to do with, sort of like the `Run()` method complains about flags that it doesn't implement? If we don't remember to do that, it'll just start quietly ignoring flags that it used to produce errors over before, and users of the `imagebuilder` CLI might come to think that those options actually work. That is an oversight that has come back to bite us before.

Yeah I was worried about imagebuilder itself now silently ignoring the flag. I don't quite have the time to actually implement the flag in this project, but at least rejecting it is necessary. In particular, it will give a dangerous false sense of security if it is silently ignored.

TomSweeneyRedHat (commented):
LGTM
and happy greeen test buttons

PR #273, opened 2024-01-09
bump github.com/containerd/containerd

nalind:
We don't use any of the code that changed to fix GHSA-7ww5-4wqc-m92c, but updating it will make dependabot feel better.


TomSweeneyRedHat:
LGTM

PR #275, opened 2024-01-11
Stages.ByName/ByTarget/ThroughTarget: accept numbers

### LINKED ISSUE #274: Generic way to build the first stage
### Discussed in https://github.com/containers/buildah/discussions/5263

<div type='discussions-op-text'>

<sup>Originally posted by **vonbox** January 11, 2024</sup>
With `buildah bud --target=layer0 .` I can build the first stage of a multistage containerfile. Is there a way to specify the first layer without me needing to name it, or without needing to know the name? Something like `buildah bud --target=0 .`?
I am trying to create a kinda generic script that should be able to build the first stage of any containerfile.</div>

nalind:
Accept integer values in ByName(), ByTarget(), and ThroughTarget() as aliases for the stages with that "Position" value.  Add a ByPosition() alongside ByName().

It would be faster to just index the stages directly instead of searching for a stage with the right Position value, but this lets us avoid a bounds check and is just easier to explain.

Fixes #274.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -198,6 +198,21 @@ func (stages Stages) ByName(name string) (Stage, bool) {
 			return stage, true
 		}
 	}
+	if i, err := strconv.Atoi(name); err == nil {
+		return stages.ByPosition(i)
+	}
+	return Stage{}, false
+}
+
+func (stages Stages) ByPosition(position int) (Stage, bool) {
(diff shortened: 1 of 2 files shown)


flouthoc (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #280, opened 2024-04-02
healthcheck: add support for `--start-interval`

flouthoc:
Add support for `--start-interval` which is present in newer version of docker
Ref: https://docs.docker.com/reference/dockerfile/#healthcheck

Needed by: https://github.com/containers/podman/issues/21701


flouthoc:

flouthoc:
If this gets merged I'll open a followup PR in buildah and vendor new imagebuilder there.

dispatchers.go
```diff
@@ -606,6 +606,7 @@ func healthcheck(b *Builder, args []string, attributes map[string]bool, flagArgs
 
 		flags := flag.NewFlagSet("", flag.ContinueOnError)
 		flags.String("start-period", "", "")
+		flags.String("start-interval", "", "")
```
nalind, line 609: The reference link specifies defaults for these settings.  Are we returning those defaults when there isn't a corresponding value supplied in the instruction?

nalind:
> If this gets merged I'll open a followup PR in buildah and vendor new imagebuilder there.

Then this should probably include a commit to bump the release version.

dispatchers.go
flouthoc: Let me add a default value here.

nalind:
> > > If this gets merged I'll open a followup PR in buildah and vendor new imagebuilder there.
> > 
> > Then this should probably include a commit to bump the release version.
> 
> @nalind Did you mean a version bump for imagebuilder itself ?

Yes, unless there's some reason not to.

TomSweeneyRedHat:
Other than @nalind 's comments, LGTM

flouthoc:
@nalind @TomSweeneyRedHat done.

dispatchers.go
nalind: Oof, I guess the real question should have been whether those defaults get baked into the image, or if they're deferred until run-time.  I _think_ a conformance test that uses a HEALTHCHECK instruction that leaves the various settings at their defaults would tell us whether or not the defaults should be provided in the Config in the committed image.

flouthoc: I have a hunch that `docker` should be doing this on runtime end and not baking into images, otherwise our previous conformance tests would have failed since we never had defaults. Let me try running the conformance test to verify this.

nalind: I'm not sure we have a conformance test for that (I see a unit test that uses Dockerfile.healthcheck and Dockerfile.badhealthcheck, but that mainly exercises the parsing logic), and if you don't see one, either, it'd probably be a good idea to add one.

nalind: Okay, good to know.  Thanks for verifying!

dockerfile/parser/testfiles/health/Dockerfile
```diff
@@ -7,4 +7,6 @@ HEALTHCHECK --interval=5s --timeout=3s --retries=3 \
 HEALTHCHECK CMD
 HEALTHCHECK   CMD   a b
 HEALTHCHECK --timeout=3s CMD ["foo"]
+HEALTHCHECK --start-interval=5s --interval=5s --timeout=3s --retries=3 \
```
nalind, line 10: This one could use a "--start-period" flag, since it's exercising the parser, and I don't see anything else in this directory checking that the flag is recognized.

nalind:
LGTM with one nit, which isn't a blocker.

dockerfile/parser/testfiles/health/Dockerfile
flouthoc: Done.

flouthoc:
@nalind Is it that CI is using older `docker` for conformance tests ? I error on travis now for the flag `--start-interval`

nalind:
> @nalind Is it that CI is using older `docker` for conformance tests ? I error on travis now for the flag `--start-interval`

It's certainly plausible.  My access is limited, so I can't actually read the error logs.

flouthoc:
@nalind  In error logs at least it is saying that docker cannot parse `start-interval` and I don't get that in my local so i guess it must be it. Do you have any clue how can i bump `docker` version in CI ?

```
=== RUN   TestConformanceInternal/healthcheck

    conformance_test.go:922: 41: unable to build Docker image "": API error (400): dockerfile parse error on line 5: unknown flag: start-interval
    ```

**Comment by nalind** [comment]
> @nalind In error logs at least it is saying that docker cannot parse `start-interval` and I don't get that in my local so i guess it must be it.

That's consistent with the version information at https://docs.travis-ci.com/user/reference/jammy/.

> Should i try this ? https://docs.travis-ci.com/user/docker/#installing-a-newer-docker-version in a different commit.

Any reason not to?

**Comment by flouthoc** [comment]
@nalind A different test is failing i.e `TestConformanceInternal/copy_chmod` , I think this is failing because changes in newer version of docker.

**Comment by flouthoc** [comment]
@nalind This fixes error for me and is compatible with newer buildkit, but I am not sure why this additional `&` was added, do you have any hints ? I can't think of a reason why it was added all the test pass without it as well.

```diff
diff --git a/dockerclient/client.go b/dockerclient/client.go
index 1e8c83f..1236d66 100644
--- a/dockerclient/client.go
+++ b/dockerclient/client.go
@@ -1058,7 +1058,7 @@ func (e *ClientExecutor) CopyContainer(container *docker.Container, excludes []s
                        }
                        chmod = func(h *tar.Header, r io.Reader) (data []byte, update bool, skip bool, err error) {
                                mode := h.Mode &^ 0o777
-                               mode |= parsed & 0o777
+                               mode |= parsed
                                h.Mode = mode
                                return nil, false, false, nil
```

nalind:
> @nalind This fixes error for me and is compatible with newer buildkit, but I am not sure why this additional `&` was added, do you have any hints ? I can't think of a reason why it was added all the test pass without it as well.
> 
> ```diff
> diff --git a/dockerclient/client.go b/dockerclient/client.go
> index 1e8c83f..1236d66 100644
> --- a/dockerclient/client.go
> +++ b/dockerclient/client.go
> @@ -1058,7 +1058,7 @@ func (e *ClientExecutor) CopyContainer(container *docker.Container, excludes []s
>                         }
>                         chmod = func(h *tar.Header, r io.Reader) (data []byte, update bool, skip bool, err error) {
>                                 mode := h.Mode &^ 0o777
> -                               mode |= parsed & 0o777
> +                               mode |= parsed
>                                 h.Mode = mode
>                                 return nil, false, false, nil
> ```

The intent there was to avoid cases where bits that attempt to change the file type were added (the bits for block and character devices fit withing 16-bit numbers).  0o7777 might work better there, since that'll also allow setuid/setgid/sticky, though I think that'd be kind of a weird thing to use in a COPY --chown.

flouthoc:
> The intent there was to avoid cases where bits that attempt to change the file type were added (the bits for block and character devices fit within g 16-bit numbers). 0o7777 might work better there, since that'll also allow setuid/setgid/sticky, though I think that'd be kind of a weird thing to use in a COPY --chown.

Thanks this worked, I think CI should pass now.

nalind:
Oof, I meant chmod, but anyway, LGTM.

flouthoc:
Okay I think all the test passed last time except one which failed and I think it was a size diff fluke in conformance test, I think it should pass in this re-push.

flouthoc:
@nalind @rhatdan @TomSweeneyRedHat PTAL , finally CI is green.

TomSweeneyRedHat:
LGTM and happy green test buttons

flouthoc:
@nalind @rhatdan PTAL

flouthoc:
@nalind @TomSweeneyRedHat Could we cut a new imagebuilder version whenever possible, thanks.

PR #281, opened 2024-04-11
Don't expand RUN heredocs ourselves, let the shell do it

nalind:
When handling RUN instructions that use heredoc syntax, don't bother interpolating environment variables and argument values, and let the command that's running handle it.  The `dockerclient` implementation doesn't support heredoc syntax, so this is being tested by a conformance test in https://github.com/containers/buildah/pull/5473.


rhatdan:
LGTM
@flouthoc PTAL

flouthoc (approved):
LGTM
/lgtm
/approve

PR #283, opened 2024-06-03
Trim trailing path separators from WORKDIR values

nalind:
Compatibility requires that we trim trailing path separators from WORKDIR values.  This should resolve https://github.com/containers/buildah/pull/5526#issuecomment-2115408076.

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -372,6 +372,10 @@ func workdir(b *Builder, args []string, attributes map[string]bool, flagArgs []s
 		workdir = filepath.Join(string(os.PathSeparator), current, workdir)
 	}
 
+	if workdir != string(os.PathSeparator) {
+		workdir = strings.TrimSuffix(workdir, string(os.PathSeparator))
+	}
+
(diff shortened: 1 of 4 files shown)


dispatchers.go
```diff
@@ -372,6 +372,10 @@ func workdir(b *Builder, args []string, attributes map[string]bool, flagArgs []s
 		workdir = filepath.Join(string(os.PathSeparator), current, workdir)
 	}
 
+	if trimmed := strings.TrimSuffix(workdir, string(os.PathSeparator)); trimmed != "" {
```
rhatdan, line 375: This code is a a little confusing, would

```
if workdir != os.PathSepator {
   workdir= strings.TrimSuffix(workdir, string(os.PathSeparator))
} 
```
Make easier to understand?

rhatdan:
LGTM other then small code suggestion.

dispatchers.go
nalind: Sure.

TomSweeneyRedHat:
LGTM

PR #286, opened 2024-06-20
dispatcher: honor heading args while processing FROM --platform

nalind:
When evaluating flag arguments to FROM instructions (currently that's just --platform=), also consider the values of ARGs defined in the heading (the part of the file before the first FROM instruction), as we do when evaluating the name of the base image or stage.


rhatdan:
LGTM
@flouthoc @giuseppe @TomSweeneyRedHat PTAL

PR #289, opened 2024-07-15
WIP: tweak ARG processing

nalind:
* Only expand ARGs in COPY instructions if they've been declared in the current stage (changed).
* Make sure that we expand the TARGET and BUILD family of built-in ARGs in FROM instructions, regardless of whether or not they're declared in the heading.
* Make sure that we only expand the TARGET and BUILD family of built-in ARGs in stages after they've been declared.
* Add some godoc for the various arg-related fields in a Builder.
* Use strings.Cut() to simplify parsing of ARG=VALUE pairs.


nalind:
Okay, I think this does what I want it to do.
/retitle Tweak ARG processing, particularly for BUILD... and TARGET...

rhatdan:
LGTM
@flouthoc PTAL

flouthoc (approved):
LGTM
/lgtm
/approve

Issue #257, opened 2023-05-30
 Build image failed without docker


zhujian7:
I have `podman` installed but without `docker`, when I execute "make images", it failed with:
```
╰─# make images
Using existing imagebuilder from /root/go/bin/imagebuilder
Warning: Installed imagebuilder version 1.2.3 does not match expected version 1.2.1.
imagebuilder --allow-pull -t quay.io/open-cluster-management/addon-manager:latest -f ./build/Dockerfile .
reading authentication configurations: open /root/.dockercfg: no such file or directory
make: *** [Makefile:50: image-addon-manager] Error 1
```

After I create the `/root/.dockercfg:` manually? I got:
```
Using existing imagebuilder from /root/go/bin/imagebuilder
Warning: Installed imagebuilder version 1.2.3 does not match expected version 1.2.1.
imagebuilder --allow-pull -t quay.io/open-cluster-management/addon-manager:latest -f ./build/Dockerfile .
Get "http://unix.sock/images/golang:1.19/json": dial unix /var/run/docker.sock: connect: no such file or directory
make: *** [Makefile:50: image-addon-manager] Error 1
```

then I followed the README to pull the image:
```
╰─# podman pull docker-daemon:golang:1.19
Error: initializing source docker-daemon:golang:1.19: loading image from docker engine: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

So want to confirm if docker is required to use imagebuilder? 

Imagebuilder version: 1.2.3

nalind:
Yes, `imagebuilder` is built to do its heavy lifting with the help of a daemon accessed through a socket at the default location, or one listed in the environment as ${DOCKER_HOST}.

PR #290, opened 2024-07-21
Ignore no docker configuration found error

zhujian7:
I am using podman, when I try to use the imagebuilder `--allow-pull` to pull a public image, I got:
```
╰─$ imagebuilder --allow-pull -t quay.io/open-cluster-management/registration:latest -f ./build/Dockerfile.registration .
reading authentication configurations: open /Users/jiazhu/.dockercfg: no such file or directory
```

After this fix, it will be:
```
╰─$ imagebuilder --allow-pull -t quay.io/open-cluster-management/registration:latest -f ./build/Dockerfile.registration .
W0721 11:34:02.229826   14874 imagebuilder.go:94] No docker configuration found
--> FROM golang:1.22-bullseye as builder
--> ARG OS=linux
--> ARG ARCH=amd64
```

related issue: https://github.com/openshift/imagebuilder/issues/257

### DIFF
--- a/cmd/imagebuilder/imagebuilder.go
+++ b/cmd/imagebuilder/imagebuilder.go
@@ -1,12 +1,14 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"log"
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/docker/distribution/reference"
 	dockerregistrytypes "github.com/docker/docker/api/types/registry"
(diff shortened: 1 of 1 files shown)


cmd/imagebuilder/imagebuilder.go
```diff
@@ -90,10 +90,14 @@ func main() {
 	options.Out, options.ErrOut = os.Stdout, os.Stderr
 	authConfigurations, err := docker.NewAuthConfigurationsFromDockerCfg()
 	if err != nil {
-		log.Fatalf("reading authentication configurations: %v", err)
+		if strings.Contains(err.Error(), "no such file or directory") {
```
rhatdan, line 93: Can you check this versus the constant rather then the string.

Something like.

if errors.Is(err, syscall.ENOENT)

zhujian7: Thanks for the review. I have fixed this, but I do not know why my changes broke the Travis CI:
```
0.16s$ docker pull busybox
Using default tag: latest
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
The command "docker pull busybox" failed and exited with 1 during .
```

rhatdan:
LGTM
@nalind PTAL

rhatdan:
@flouthoc PTAL

cmd/imagebuilder/imagebuilder.go
nalind: You didn't break it.  I'm not sure why it worked before, but there was no point where the Travis configuration was starting the daemon after stopping it and attempting to update it.  I've opened https://github.com/openshift/imagebuilder/pull/291 to fix that.  You can cherry-pick it onto your PR if you like, or rebase after it's merged to pick it up.

zhujian7: Thanks @nalind, I will rebase after it's merged.

flouthoc (approved):
LGTM

PR #292, opened 2024-07-22
Tag v1.2.12

nalind:
Tag v.1.2.12 for a release.


nalind:
* Changed `arg()` behavior in accordance with comments at https://github.com/containers/buildah/blob/16bd9a31054dfb6b14fc4c9423ef1a7a7d81aef0/imagebuildah/build.go#L418.
* Copy the new `BuiltinArgDefaults` field's contents from the main `Builder` into the per-stage copy in `NewStages()`, which I forgot to do in #289.

PR #293, opened 2024-07-22
NewStages(): resolve stage names using ARGs

nalind:
Resolve stage names using header and built-in args.


rhatdan:
LGTM
@flouthoc PTAL

flouthoc (approved):
LGTM

PR #294, opened 2024-07-23
TestDispatchArgTargetPlatformGood(): ensure that TARGETOS can be set

nalind:
Ensure that TARGETOS, TARGETARCH, and TARGETVARIANT can be overridden.


nalind:
I'm going to want to tag a 1.2.13 with this, to undo breaking this in 1.2.12.

PR #301, opened 2025-02-19
Set empty `TARGETVARIANT` and `BUILDVARIANT`

Honny1:
This PR sets empty values as the default value for the `TARGETVARIANT` and `BUILDVARIANT` buildin arguments since Docker does not insert these values for `ARM64`.

Related PR:
- https://github.com/containers/buildah/pull/5990

### DIFF
--- a/dispatchers.go
+++ b/dispatchers.go
@@ -41,18 +41,11 @@ var builtinArgDefaults = map[string]string{
 	"TARGETPLATFORM": localspec.OS + "/" + localspec.Architecture,
 	"TARGETOS":       localspec.OS,
 	"TARGETARCH":     localspec.Architecture,
-	"TARGETVARIANT":  localspec.Variant,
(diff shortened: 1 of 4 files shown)


nalind:
I'm seeing errors from unit tests in the main directory when I run them in my arm64-on-amd64 VM with these changes.  Are you getting passing results when running `go test ./...` at the top level?

Honny1:
I also failed unit tests. (Including master) I checked the failed tests and they expect that a variant should be set. I think the tests are wrong. WDYT?

nalind:
Yes, if they expected behavior which we now consider to be incorrect, they need to be updated to expect what we consider to be correct.

Honny1:
I adjusted the tests to expect the correct values.

nalind:
/lgtm
Thanks!

PR #303, opened 2025-02-19
Add arm64 in Travis

nalind:
Start conformance tests in CI on both amd64 and arm64


nalind:
@TomSweeneyRedHat @rhatdan @mrunalp PTAL

PR #306, opened 2025-08-20
Use go.podman.io/storage from monorepo instead of github.com/containers/storage.

jankaluza:
This commit switch from github.com/containers/storage to go.podman.io/storage.

It has been generated by:

```
$ vi go.mod to change the dependency
$ find . -type f -name '*.go' -exec sed -i -e 's,"github.com/containers/storage,"go.podman.io/storage,g' {} \;
$ goimports -v -w .
$ go mod tidy
$ go mod vendor
```


go.mod
```diff
@@ -1,16 +1,16 @@
 module github.com/openshift/imagebuilder
 
-go 1.22.0
+go 1.23.3
```
TomSweeneyRedHat, line 3: @nalind or @flouthoc any concerns with this Go bump?

nalind: The openshift/builder repository is using "toolchain 1.23.1", and containers projects' main branches are on 1.23, so I think we're okay.

TomSweeneyRedHat:
LGTM
once the tests are hip, and if @nalind and @flouthoc don't have issues with the Go bump

nalind:
What's with the whitespace changes in `evaluator.go`?

jankaluza:
@nalind , it's like `goimports -v -w .` formats the file. I wanted to change it, but also vscode auto-formats this by default the very same way, so I thought it is actually the preferred way for go.

I understand this is not directly related to the code-change I'm doing here, so I'm OK to remove these white-space changes if you think so. Just tell me.

PR #324, opened 2026-08-16
Allow symbolic modes in COPY/ADD --chmod

### LINKED ISSUE #6938: COPY/ADD --chmod rejects symbolic modes (only octal accepted) — BuildKit parity gap
## Summary

buildah's `COPY --chmod` / `ADD --chmod` accepts **octal numeric modes only**. It rejects all **symbolic** modes (`u+x`, `go-w`, `a+rX`, `a+rX,go-w`, …) and the Go `0o755` prefix, all of which BuildKit accepts. This breaks building Dockerfiles that use BuildKit-style symbolic `--chmod` — an increasingly common pattern.

## Empirical results

Minimal repro:
```
printf 'FROM scratch\nCOPY --chmod=<MODE> ctx.txt /ctx.txt\n' > Dockerfile && podman build .
```

buildah `1.44.0` (podman 6.0.0 server):

| `--chmod=` | result |
|---|---|
| `0755`, `0644` | ✅ accepted |
| `0o755` | ❌ rejected |
| `a+rx`, `a+rwx` | ❌ rejected |
| `u+x` | ❌ rejected |
| `go-w` | ❌ rejected |
| `a+rX` | ❌ rejected |
| `a+rX,go-w`, `u+rX,go-w` | ❌ rejected |

Only plain octal (no `0o` prefix) is accepted. Every symbolic form — including the simple `u+x` — fails during `COPY`/`ADD` with `Error parsing chmod <mode>`.

## Impact

Dockerfiles using BuildKit-style symbolic `--chmod` cannot be built with buildah/podman. Real example: `COPY --link --chmod=a+rX,go-w . .`, which projects use to set "readable + traversable by non-root, no group/other write" in one layer instead of a separate `chmod -R` pass over a large tree.

The workaround is to rewrite the mode to octal, but that (a) must be reapplied per build, and (b) is not always a 1:1 conversion — symbolic modes like `a+rX` are *conditional* on each file's existing mode (capital `X` = "execute only for dirs / already-executable files"), so a fixed octal constant is a lossy approximation.

## Request

Support symbolic `--chmod` modes in `COPY`/`ADD` for BuildKit parity, resolving each clause against the copied file's current mode (as BuildKit does). Even partial support for the common `a`/`u`/`g`/`o` ± `r`/`w`/`x`/`X` forms would unblock most real-world Dockerfiles.

## Environment

- buildah `1.44.0` (podman `6.0.0` server, macOS applehv machine)
- BuildKit (Docker buildx) accepts all the rejected forms

paoloantinori:
Fixes the Containerfile `COPY/ADD --chmod=` instruction path to accept symbolic modes, restoring parity with the dockerfile frontend spec (numeric and symbolic `--chmod` are both part of the spec since Dockerfile syntax 1.14) and with `buildah copy/add --chmod=`, which already resolve symbolic values.

## Problem

`checkChmodConversion` validated `--chmod=` with `strconv.ParseUint` (octal only) at dispatch time, so every symbolic form (`+x`, `u+x`, `a+rX,go-w`, ...) failed with `Error parsing chmod <mode>` before the executor ever saw it. This is the root cause of podman-container-tools/buildah#6938: buildah's execution layer is already symbolic-aware (PR buildah#6778), but the parser vendored from here rejected the instruction form. That issue includes a repro on released buildah 1.45.0 showing the CLI subcommand and the Containerfile instruction disagreeing on the very same binary.

## Change

- `checkChmodConversion` accepts octal (now range-checked to 0-07777, as BuildKit's frontend does) and validates symbolic modes by delegating to `mode.Parse` from `github.com/tonistiigi/dchapes-mode`. Delegation rather than a local grammar re-implementation: chmod(1) allows repeated op-perms groups within a clause (`u+r-w`), which a hand-rolled subset validator gets wrong; dchapes-mode is the library buildah's executor uses to resolve these values and what BuildKit's frontend uses. `Parse` only builds the command list (no base mode needed); resolution against each file's current mode, where the conditional `X` is meaningful, stays with the executor.
- The `dockerclient` executor resolved `Copy.Chmod` with `strconv.ParseInt` octal-only; with the parser accepting symbolic values it would have failed mid-execution with a fuzzier error. It now parses once with `mode.Parse` and applies per file via the new `applyChmod` helper, which goes through `h.FileInfo().Mode()` (so symbolic clauses see the directory type and special bits correctly) and maps setuid/setgid/sticky back onto their unix bit positions in the tar header. A raw cast of the tar mode to `os.FileMode` would write those bits in the wrong positions and silently drop them. `applyChmod` has unit coverage (numeric, symbolic, special bits, directory-X); the conformance suite only exercises this behind a build tag plus a live daemon.
- Documented the `Copy.Chmod` field contract (octal 0-07777 or chmod(1) symbolic clauses, passed to the executor for resolution), mirroring `Chown`.

The new dependency is stdlib-only, two files, already in the module graph at this exact pseudo-version: `moby/buildkit` v0.29.0 (a direct dependency) requires it, so buildah/podman consumers see no new modules.

## Behavior

- Octal: unchanged apart from two deliberate alignments with chmod(1) and BuildKit, both called out below.
- Symbolic clause lists pass through verbatim in `Copy.Chmod` for executors to resolve.

Deliberate behavior changes (reviewers please sanity-check):

1. Octal values above 07777 (e.g. `17777`) are now rejected at parse time instead of being masked down at execution.
2. In the `dockerclient` executor, numeric (absolute) modes now clear pre-existing setuid/setgid/sticky bits instead of preserving them, matching `chmod(1)` absolute-mode semantics.

Verified end-to-end against buildah `main` (replace directive to this branch; `FROM scratch` builds; modes read back from the built image layer):

| source mode | `--chmod=` | result |
|---|---|---|
| 700 | `a+rX,go-w` | 755 |
| 644 | `+x` | 755 |
| 644 | `u+x,go-w` | 744 |
| 700 | `u+rX-w` | 500 |
| 644 | `g+r=rx` | 654 |

The conditional `X` resolves per-file as expected (`a+rX` on a 0700 script yields 0755). Octal builds are unchanged, and `888`, `rwxrwxrwx`, `0o755`, and `17777` are rejected.

Fixes podman-container-tools/buildah#6938 (see also podman-container-tools/podman#28293). Once this lands, buildah needs only a revendor (the `Copy.Chmod` to `AddAndCopyOptions.Chmod` wiring already exists on its `main`); I can send that PR with a `bud` test as soon as there is a revendorable commit.


dispatchers.go
```diff
@@ -814,10 +815,19 @@ func shell(b *Builder, args []string, attributes map[string]bool, flagArgs []str
...
-	_, err := strconv.ParseUint(chmod, 8, 32)
-	if err != nil {
+	if v, err := strconv.ParseUint(chmod, 8, 32); err == nil {
+		if v > 0o7777 {
+			return fmt.Errorf("Error parsing chmod %s: it should be octal and between 0 and 07777", chmod)
+		}
+		return nil
+	}
+	if _, err := mode.Parse(chmod); err != nil {
 		return fmt.Errorf("Error parsing chmod %s", chmod)
```
TomSweeneyRedHat: Would be good to add the err here
```suggestion
		return fmt.Errorf("Error parsing chmod %s: %w", chmod, err)
```

dockerclient/client.go
```diff
@@ -1069,14 +1091,12 @@ func (e *ClientExecutor) CopyContainer(container *docker.Container, excludes []s
 	for _, c := range copies {
 		var chmod func(h *tar.Header, r io.Reader) (data []byte, update bool, skip bool, err error)
 		if c.Chmod != "" {
-			parsed, err := strconv.ParseInt(c.Chmod, 8, 16)
+			parsed, err := mode.Parse(c.Chmod)
 			if err != nil {
-				return err
+				return fmt.Errorf("invalid chmod %q", c.Chmod)
```
TomSweeneyRedHat: ditto error add here
```suggestion
				return fmt.Errorf("invalid chmod %q : %w", c.Chmod, err)
```

TomSweeneyRedHat:
Nice change overall.  LGTM
Just a few nit suggestions for consideration

dockerclient/client.go
paoloantinori: Done in 2ca5e05, thanks!

paoloantinori:
Both nits addressed in 2ca5e05 (error now wraps the underlying parse error in both spots). Tests re-run: root, dockerclient, parser and imageprogress packages all ok.

dispatchers.go
paoloantinori: Done in 2ca5e05, thanks!

TomSweeneyRedHat:
Changes LGTM
Tests are still running though, and I didn't expect that.  @nalind?

nalind:
Might be related to github's issues yesterday.

/ok-to-test

I restarted the Travis tests, and they appear to be failing in a conformance test that needs [TestConformanceExternal/copy and env interaction: update test context](https://github.com/openshift/imagebuilder/pull/325/commits/3aae86fb19896816b16bb3c7c894df7c82cafefb) from #325 to catch it up with some changes in the repository that it tests building from.

@paoloantinori, if you want to cherry-pick that to get the conformance tests to pass here to not be blocked on that one getting reviewed and merged, that should sort this one out.

paoloantinori:
Thanks both for the reviews and the CI restarts. I'll confess I'm not fully familiar with this project's merge process, so I hope I'm not being pushy or bothering anyone with the following - please just ignore this if there's a queue I should be waiting in.

Since the text LGTMs don't seem to set the label on their own: @TomSweeneyRedHat, would you mind dropping a formal `/lgtm` if you're still comfortable with the change? Per the approval bot's note, I'm also assigning @nalind for the final approval.

/assign @nalind

dispatchers.go
paoloantinori: Addressed in 2ca5e05: the wrap now happens inside checkChmodConversion itself (dispatchers.go:831, "Error parsing chmod %s: %w"), so both dispatch call sites get the chained error from a single wrap point.

dockerclient/client.go
paoloantinori: Addressed in 2ca5e05: dockerclient/client.go:1096 now wraps it as "invalid chmod %q: %w".

TomSweeneyRedHat:
LGTM
@paoloantinori Thanks for hopping through the hoops!
@nalind?

paoloantinori:
Thanks @TomSweeneyRedHat! One small note: the approval bot only registers the command when it sits on its own line, so the LGTM in your comment above hasn't set the label yet. Whenever convenient, a bare "/lgtm" line (backticked or alone, without other text around it) would do it. Not rushing — just flagging it since the process is opaque from the outside.

nalind (approved):
LGTM, thanks!
/lgtm

